PR #1, opened 2022-05-11
Initial implementation

michaelsbradleyjr:
## TODO

All completed items include tests providing good code coverage.

- [x] `Namespace` and `Key`
- [x] base type and methods: `Datastore`, `contains`, `delete`, `get`, `put`
- [x] `NullDatastore`, `FilesystemDatastore`, `SQLiteDatastore`, `TieredDatastore`
- [ ] `LRUCacheDatastore`
- [ ] `Query` type and `method query` for `Datastore` (base and impls)
- [ ] Lifting of methods and helpers so it's more convenient to work with e.g. `?!Key`, `?!Datastore`
- [ ] Inline docs and examples of usage in the readme
- [ ] Split some `Datastore` impls into their own repositories, e.g. `SQLiteDatastore` could be hosted in `status-im/nim-sqlite-datastore`

Uncompleted items may be addressed in subsequent PRs and will benefit from feeback on this initial implementation. It may be desirable that some are completed prior to merging this PR.

## Goals

This initial implementation aims to produce a library that can be experimented with and used in the context of ongoing and new projects, e.g. [nim-codex](https://github.com/status-im/nim-codex). It should therefore be fairly minimal while providing the overall shape and direction of this library's future.

## Overview

nim-datastore is inspired by the Python library [datastore](https://github.com/datastore/datastore), which has a number of subprojects e.g. [datastore-mongo](https://github.com/datastore/datastore.mongo) that implement the datastore concepts for various backends.

Work on [Python datastore](https://github.com/datastore/datastore) stopped some years ago, but there are actively maintained ports for other languages, e.g. [go-datastore](https://github.com/ipfs/go-datastore) and [interface-datastore](https://github.com/ipfs/js-ipfs-interfaces/tree/master/packages/interface-datastore) (JavaScript).

From the [go-datastore](https://github.com/ipfs/go-datastore) readme:
> Datastore is a generic layer of abstraction for data store and database access. It is a simple API with the aim to enable application development in a datastore-agnostic way, allowing datastores to be swapped seamlessly without changing application code. Thus, one can leverage different datastores with different strengths without committing the application to one datastore throughout its lifetime.
>
> In addition, grouped datastores significantly simplify interesting data access patterns (such as caching and sharding).

From the [Python datastore](https://github.com/datastore/datastore) readme:
> Datastore implementors are free to implement specialized features ... [but] remember the goal of this project: simple, unified API for multiple data stores. When making heavy use of a particular library's specific functionality, perhaps one should not use datastore and should directly use that library.

## This implementation

The current work on nim-datastore has been informed by the [Python datastore](https://github.com/datastore/datastore) codebase while also referencing the ports (and their subprojects) linked above.

Nim's [methods](https://nim-lang.org/docs/manual.html#methods) facility has been used to define a base `Datastore` type and its API.

Because Nim is a statically typed language, and because generic methods and multi-methods are deprecated, concrete choices have been made re: `key` and `data` parameters:
* `key` corresponds to a `Key` type, included in this implementation along with `Namespace`, using `string` and a specific set of rules for how strings are joined/split into instances of `Namespace` and `Key`.
* `data` corresponds to `openArray[byte]` (and returned data to `seq[byte]`), leaving it entirely to the user to decide how to de/serialize inputs and outputs.

There are upsides and downsides to this approach, and it also has implications for the to-be-implemented `Query` type and `method query`.

A more flexible solution might involve a set of [generics](https://nim-lang.org/docs/manual.html#generics)-based forward declarations that are `include`d by implementations of `Datastore`. Or maybe there's a better alternative.

An issue will be created for the purpose of discussing the shape and direction of `Query` / `method query` (this PR description will be updated with a link to it).

---

`[wip]` commits in this PR branch will be squashed pending review and approval of team members.

### DIFF
--- a/datastore/filesystem_datastore.nim
+++ b/datastore/filesystem_datastore.nim
@@ -1,0 +1,159 @@
+import std/os
+
+import pkg/questionable
+import pkg/questionable/results
+import pkg/upraises
+
+import ./datastore
+
+export datastore
+
+push: {.upraises: [].}
+
+type
+  FileSystemDatastore* = ref object of Datastore
+    root: string
+
+const
+  objExt* = ".dsobject"
+
+proc new*(
+  T: type FileSystemDatastore,
+  root = "data"): ?!T =
+
+  try:
+    let
+      root = if root.isAbsolute: root
+             else: getCurrentDir() / root
+
+    createDir(root)
+    success T(root: root)
+
+  except IOError as e:
+    failure e
+
+  except OSError as e:
+    failure e
+
+proc root*(self: FileSystemDatastore): string =
+  self.root
+
+proc path*(
+  self: FileSystemDatastore,
+  key: Key): string =
+
+  var
+    segments: seq[string]
+
+  for ns in key:
+    without field =? ns.field:
+      segments.add ns.value
+      continue
+
+    segments.add(field / ns.value)
+
+  # is it problematic that per this logic Key(/a:b) evaluates to the same path
+  # as Key(/a/b)? may need to check if/how other Datastore implementations
+  # distinguish them
+
+  self.root / joinPath(segments) & objExt
+
+method contains*(
+  self: FileSystemDatastore,
+  key: Key): ?!bool {.locks: "unknown".} =
+
+  success fileExists(self.path(key))
+
+method delete*(
+  self: FileSystemDatastore,
+  key: Key): ?!void {.locks: "unknown".} =
+
+  let
+    path = self.path(key)
+
+  try:
+    removeFile(path)
+    success()
+
+    # removing an empty directory might lead to surprising behavior depending
+    # on what the user specified as the `root` of the FileSystemDatastore, so
+    # until further consideration, empty directories will be left in place
+
+  except OSError as e:
+    failure e
+
+method get*(
+  self: FileSystemDatastore,
+  key: Key): ?!(?seq[byte]) {.locks: "unknown".} =
+
+  # to support finer control of memory allocation, maybe could/should change
+  # the signature of `get` so that it has a 3rd parameter
+  # `bytes: var openArray[byte]` and return type `?!bool`; this variant with
+  # return type `?!(?seq[byte])` would be a special case (convenience method)
+  # calling the former after allocating a seq with size automatically
+  # determined via `getFileSize`
+
+  let
+    path = self.path(key)
+    exists = ? self.contains(key)
+
+  if exists:
+    var
+      file: File
+
+    if not file.open(path):
+      return failure "unable to open file: " & path
+    else:
+      try:
+        let
+          size = file.getFileSize
+
+        var
+          bytes: seq[byte]
+
+        if size > 0:
+          newSeq(bytes, size)
+
+          let
+            bytesRead = file.readBytes(bytes, 0, size)
+
+          if bytesRead < size:
+            return failure $bytesRead & " bytes were read from " & path &
+              " but " & $size & " bytes were expected"
+
+        success bytes.some
+
+      except IOError as e:
+        failure e
+
+      finally:
+        file.close
+
+  else:
+    success seq[byte].none
+
+method put*(
+  self: FileSystemDatastore,
+  key: Key,
+  data: openArray[byte]): ?!void {.locks: "unknown".} =
+
+  let
+    path = self.path(key)
+
+  try:
+    createDir(parentDir(path))
+    if data.len > 0: writeFile(path, data)
+    else: writeFile(path, "")
+    success()
+
+  except IOError as e:
+    failure e
+
+  except OSError as e:
+    failure e
+
+# method query*(
+#   self: FileSystemDatastore,
+#   query: ...): ?!(?...) {.locks: "unknown".} =
+#
+#   success ....none
--- a/datastore/key.nim
+++ b/datastore/key.nim
@@ -1,0 +1,375 @@
+import std/algorithm
+import std/hashes
+import std/oids
+import std/sequtils
+import std/strutils
+
+import pkg/questionable
+import pkg/questionable/results
+import pkg/upraises
+
+export hashes
+
+push: {.upraises: [].}
+
+type
+  Namespace* = object
+    field: ?string
+    value: string
+
+  Key* = object
+    namespaces: seq[Namespace]
+
+const
+  delimiter = ":"
+  separator = "/"
+
+# TODO: operator/s for combining string|Namespace,string|Namespace
+# TODO: lifting from ?![Namespace|Key] for various ops
+
+proc init*(
+  T: type Namespace,
+  field, value: string): ?!T =
+
+  if value == "":
+    return failure "value string must not be empty"
+
+  if value.strip == "":
+    return failure "value string must not be all whitespace"
+
+  if value.contains(delimiter):
+    return failure "value string must not contain delimiter \"" &
+      delimiter & "\""
+
+  if value.contains(separator):
+    return failure "value string must not contain separator \"" &
+      separator & "\""
+
+  if field != "":
+    if field.strip == "":
+      return failure "field string must not be all whitespace"
+
+    if field.contains(delimiter):
+      return failure "field string must not contain delimiter \"" &
+        delimiter & "\""
+
+    if field.contains(separator):
+      return failure "field string must not contain separator \"" &
+        separator & "\""
+
+    success T(field: field.some, value: value)
+  else:
+    success T(field: string.none, value: value)
+
+proc init*(
+  T: type Namespace,
+  id: string): ?!T =
+
+  if id == "":
+    return failure "id string must not be empty"
+
+  if id.strip == "":
+    return failure "id string must not be all whitespace"
+
+  if id.contains(separator):
+    return failure "id string must not contain separator \"" & separator & "\""
+
+  if id == delimiter:
+    return failure "value in id string \"[field]" & delimiter &
+      "[value]\" must not be empty"
+
+  let
+    s = id.split(delimiter)
+
+  if s.len > 2:
+    return failure "id string must not contain more than one delimiter \"" &
+      delimiter & "\""
+
+  var
+    field: ?string
+    value: string
+
+  if s.len == 1:
+    value = s[0]
+  else:
+    value = s[1]
+
+    if value == "":
+      return failure "value in id string \"[field]" & delimiter &
+        "[value]\" must not be empty"
+
+    if value.strip == "":
+      return failure "value in id string \"[field]" & delimiter &
+        "[value]\" must not be all whitespace"
+
+    else:
+      let
+        f = s[0]
+
+      if f != "":
+        if f.strip == "":
+          return failure "field in id string \"[field]" & delimiter &
+            "[value]\" must not be all whitespace"
+
+        else:
+          field = f.some
+
+  success T(field: field, value: value)
+
+proc value*(self: Namespace): string =
+  self.value
+
+proc field*(self: Namespace): ?string =
+  self.field
+
+proc `type`*(self: Namespace): ?string =
+  self.field
+
+proc kind*(self: Namespace): ?string =
+  self.`type`
+
+proc id*(self: Namespace): string =
+  if field =? self.field: field & delimiter & self.value
+  else: self.value
+
+proc `$`*(namespace: Namespace): string =
+  "Namespace(" & namespace.id & ")"
+
+proc init*(
+  T: type Key,
+  namespaces: varargs[Namespace]): ?!T =
+
+  if namespaces.len == 0:
+    failure "namespaces must contain at least one Namespace"
+  else:
+    success T(namespaces: @namespaces)
+
+proc init*(
+  T: type Key,
+  namespaces: varargs[string]): ?!T =
+
+  if namespaces.len == 0:
+    failure "namespaces must contain at least one Namespace id string"
+  else:
+    var
+      nss: seq[Namespace]
+
+    for s in namespaces:
+      let
+        nsRes = Namespace.init(s)
+      # if `without ns =? Namespace.init(s), e:` is used `e` is nil in the body
+      # at runtime, why?
+      without ns =? nsRes:
+        return failure "namespaces contains an invalid Namespace: " &
+          nsRes.error.msg
+
+      nss.add ns
+
+    success T(namespaces: nss)
+
+proc init*(
+  T: type Key,
+  id: string): ?!T =
+
+  if id == "":
+    return failure "id string must contain at least one Namespace"
+
+  if id.strip == "":
+    return failure "id string must not be all whitespace"
+
+  let
+    nsStrs = id.split(separator).filterIt(it != "")
+
+  if nsStrs.len == 0:
+    return failure "id string must not contain only one or more separator " &
+      "\"" & separator & "\""
+
+  let
+    keyRes = Key.init(nsStrs)
+  # if `without key =? Key.init(nsStrs), e:` is used `e` is nil in the body
+  # at runtime, why?
+  without key =? keyRes:
+    return failure "id string contains an invalid Namespace:" &
+      keyRes.error.msg.split(":")[1..^1].join("").replace("\"\"", "\":\"")
+
+  success key
+
+proc namespaces*(self: Key): seq[Namespace] =
+  self.namespaces
+
+proc list*(self: Key): seq[Namespace] =
+  self.namespaces
+
+proc random*(T: type Key): string =
+  $genOid()
+
+template `[]`*(
+  key: Key,
+  x: auto): auto =
+
+  key.namespaces[x]
+
+proc len*(self: Key): int =
+  self.namespaces.len
+
+iterator items*(key: Key): Namespace {.inline.} =
+  var
+    i = 0
+
+  while i < key.len:
+    yield key[i]
+    inc i
+
+proc reversed*(self: Key): Key =
+  Key(namespaces: self.namespaces.reversed)
+
+proc reverse*(self: Key): Key =
+  self.reversed
+
+proc name*(self: Key): string =
+  self[^1].value
+
+proc `type`*(self: Key): ?string =
+  self[^1].field
+
+proc kind*(self: Key): ?string =
+  self.`type`
+
+proc instance*(
+  self: Key,
+  value: Namespace): Key =
+
+  let
+    last = self[^1]
+
+    inst =
+      if last.field.isSome:
+        @[Namespace(field: last.field, value: value.value)]
+      else:
+        @[Namespace(field: last.value.some, value: value.value)]
+
+    namespaces =
+      if self.namespaces.len == 1:
+        inst
+      else:
+        self.namespaces[0..^2] & inst
+
+  Key(namespaces: namespaces)
+
+proc instance*(self, value: Key): Key =
+  self.instance(value[^1])
+
+proc instance*(self, value: Namespace): Key =
+  Key(namespaces: @[self]).instance(value)
+
+proc instance*(
+  self: Namespace,
+  value: Key): Key =
+
+  self.instance(value[^1])
+
+proc instance*(
+  self: Key,
+  id: string): ?!Key =
+
+  without key =? Key.init(id), e:
+    return failure e
+
+  success self.instance(key)
+
+proc isTopLevel*(self: Key): bool =
+  self.len == 1
+
+proc parent*(self: Key): ?!Key =
+  if self.isTopLevel:
+    failure "key has no parent"
+  else:
+    success Key(namespaces: self.namespaces[0..^2])
+
+proc parent*(self: ?!Key): ?!Key =
+  let
+    key = ? self
+
+  key.parent
+
+proc path*(self: Key): ?!Key =
+  let
+    parent = ? self.parent
+
+  without kind =? self[^1].kind:
+    return success parent
+
+  success Key(namespaces: parent.namespaces & @[Namespace(value: kind)])
+
+proc path*(self: ?!Key): ?!Key =
+  let
+    key = ? self
+
+  key.path
+
+proc child*(
+  self: Key,
+  ns: Namespace): Key =
+
+  Key(namespaces: self.namespaces & @[ns])
+
+proc `/`*(
+  self: Key,
+  ns: Namespace): Key =
+
+  self.child(ns)
+
+proc child*(
+  self: Key,
+  namespaces: varargs[Namespace]): Key =
+
+  Key(namespaces: self.namespaces & @namespaces)
+
+proc child*(self, key: Key): Key =
+  Key(namespaces: self.namespaces & key.namespaces)
+
+proc `/`*(self, key: Key): Key =
+  self.child(key)
+
+proc child*(
+  self: Key,
+  keys: varargs[Key]): Key =
+
+  Key(namespaces: self.namespaces & concat(keys.mapIt(it.namespaces)))
+
+proc child*(
+  self: Key,
+  ids: varargs[string]): ?!Key =
+
+  let
+    ids = ids.filterIt(it != "")
+
+  var
+    keys: seq[Key]
+
+  for id in ids:
+    let
+      key = ? Key.init(id)
+
+    keys.add key
+
+  success self.child(keys)
+
+proc `/`*(
+  self: Key,
+  id: string): ?!Key =
+
+  self.child(id)
+
+proc isAncestorOf*(self, other: Key): bool =
+  if other.len <= self.len: false
+  else: other.namespaces[0..<self.len] == self.namespaces
+
+proc isDescendantOf*(self, other: Key): bool =
+  other.isAncestorOf(self)
+
+proc id*(self: Key): string =
+  separator & self.namespaces.mapIt(it.id).join(separator)
+
+proc `$`*(key: Key): string =
+  "Key(" & key.id & ")"
--- a/datastore/sqlite.nim
+++ b/datastore/sqlite.nim
@@ -1,0 +1,226 @@
+import pkg/questionable
+import pkg/questionable/results
+import pkg/sqlite3_abi
+import pkg/upraises
+
+push: {.upraises: [].}
+
+# Adapted from:
+# https://github.com/status-im/nwaku/blob/master/waku/v2/node/storage/sqlite.nim
+
+type
+  AutoDisposed*[T: ptr|ref] = object
+    val*: T
+
+  DataProc* = proc(s: RawStmtPtr) {.closure.}
+
+  NoParams* = tuple # empty tuple
+
+  NoParamsStmt* = SQLiteStmt[NoParams, void]
+
+  RawStmtPtr* = ptr sqlite3_stmt
+
+  SQLite* = ptr sqlite3
+
+  SQLiteStmt*[Params, Res] = distinct RawStmtPtr
+
+proc bindParam(
+  s: RawStmtPtr,
+  n: int,
+  val: auto): cint =
+
+  when val is openarray[byte]|seq[byte]:
+    if val.len > 0:
+      # `SQLITE_TRANSIENT` "indicate[s] that the object is to be copied prior
+      # to the return from sqlite3_bind_*(). The object and pointer to it
+      # must remain valid until then. SQLite will then manage the lifetime of
+      # its private copy."
+      sqlite3_bind_blob(s, n.cint, unsafeAddr val[0], val.len.cint,
+        SQLITE_TRANSIENT)
+    else:
+      sqlite3_bind_blob(s, n.cint, nil, 0.cint, nil)
+  elif val is int32:
+    sqlite3_bind_int(s, n.cint, val)
+  elif val is uint32:
+    sqlite3_bind_int(s, n.cint, int(val).cint)
+  elif val is int64:
+    sqlite3_bind_int64(s, n.cint, val)
+  elif val is float64:
+    sqlite3_bind_double(s, n.cint, val)
+  elif val is string:
+    # `-1` implies string length is num bytes up to first null-terminator;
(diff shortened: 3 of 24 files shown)


datastore/key.nim
```diff
@@ -0,0 +1,111 @@
...
+
+type
+  Namespace* = object
+    field: ?string
+    value: string
+
+  Key* = object
+    namespaces: seq[Namespace]
+
+const
```
dryajov, line 23: maybe, this can be re-defined as operators if it makes sense

michaelsbradleyjr: Sorry, missed this comment earlier. Yes, operators do make sense, at least an operator for building a composite `Key` makes sense.

See `` proc `/`* ``:
* https://github.com/status-im/nim-datastore/blob/initial_impl/datastore/key.nim#L288-L290
* https://github.com/status-im/nim-datastore/blob/initial_impl/datastore/key.nim#L303-L304
* https://github.com/status-im/nim-datastore/blob/initial_impl/datastore/key.nim#L330-L332

I have an idea for an operator/s re: `Namespace` (for building namespaces from strings or mixed string/Namespace args) and played with it earlier, but it wasn't clear to me how useful it would be versus aliasing `Namespace.init` to `ns` (e.g. `"a:b".ns`). I decided it's probably best to leave that until the PR is farther along.

datastore/key.nim
```diff
@@ -0,0 +1,347 @@
...
+      nss: seq[Namespace]
+
+    for s in namespaces:
+      let
+        nsRes = Namespace.init(s)
+      # if `without ns =? Namespace.init(s), e:` is used `e` is nil in the body
+      # at runtime, why?
+      without ns =? nsRes:
+        return failure "namespaces contains an invalid Namespace: " &
+          nsRes.error.msg
```
michaelsbradleyjr, line 164: @markspanbroek I'm not sure if I discovered a bug/limitation of `questionable` or my usage is incorrect.

See also L159-L165 below.

Menduist: Seems to be https://github.com/status-im/questionable/issues/11

datastore/filesystem_datastore.nim
```diff
@@ -0,0 +1,159 @@
...
+    segments: seq[string]
+
+  for ns in key:
+    without field =? ns.field:
+      segments.add ns.value
+      continue
+
+    segments.add(field / ns.value)
+
+  # is it problematic that per this logic Key(/a:b) evaluates to the same path
```
dryajov, line 55: I say that in the absense of a clear spec arround some of of the behaviours, we should match what other implementations are doing.

datastore/filesystem_datastore.nim
```diff
@@ -0,0 +1,159 @@
...
+  key: Key): ?!void {.locks: "unknown".} =
+
+  let
+    path = self.path(key)
+
+  try:
+    removeFile(path)
+    success()
+
+    # removing an empty directory might lead to surprising behavior depending
```
dryajov, line 78: This is extremelly dangerous functionality in a filesystem module in general. We should make sure that this doesn't allow blowing up `/` or some other important directory, perhaps it makes sense to have hardcoded list of common "protected" directories that the user can delete only if setting a flag manually.

datastore/filesystem_datastore.nim
michaelsbradleyjr: It seems to vary by implementation as well. While working on a first-pass at `method query` I got an idea re: how to clear this up, will take a stab at it and ask for feedback re: whether it's the behavior we want.

datastore/filesystem_datastore.nim
michaelsbradleyjr: At present it will only attempt to remove a single file corresponding to a single key.

Per `Key.init`, a key cannot be empty, so even if the `root` is `/`, then `self.path(key)` would always evaluate to something like `/a.obj`.

**NOTE**: looking carefully again now, I do think I should check that if `strip` is applied to field or value of a Namespace and it results in an empty string, then that should be an error. I'll make the change and add corresponding tests.

`proc removeFile(file: string)` of `std/os` will raise an `OSError` if `file` is actually a path to a directory (empty or otherwise). I just tried it on Linux and macOS and that was the observed behavior. Looking at the [source](https://github.com/nim-lang/Nim/blob/version-1-6/lib/pure/os.nim#L1939-L1986) and tracing back to use of Windows' `deleteFileW` / `deleteFileA`, the behavior should be the same on Windows as well because those functions can't be used to delete a directory.

datastore/filesystem_datastore.nim
```diff
@@ -0,0 +1,159 @@
...
+export datastore
+
+push: {.upraises: [].}
+
+type
+  FileSystemDatastore* = ref object of Datastore
+    root: string
+
+const
+  objExt* = ".obj"
```
Bulat-Ziganshin: .obj files are already used in Windows for compiled code (similar to .o in Unix). May be use another extension, such as .dsobject?

michaelsbradleyjr: Ah, great tip, didn't know that.

`.dsobject` works for me.

michaelsbradleyjr: Change made.

datastore/sqlite.nim
```diff
@@ -0,0 +1,238 @@
...
+  when val is openarray[byte]|seq[byte]:
+    if val.len > 0:
+      try:
+        # `SQLITE_TRANSIENT` "indicate[s] that the object is to be copied prior
+        # to the return from sqlite3_bind_*(). The object and pointer to it
+        # must remain valid until then. SQLite will then manage the lifetime of
+        # its private copy."
+        sqlite3_bind_blob(s, n.cint, unsafeAddr val[0], val.len.cint,
+          SQLITE_TRANSIENT)
+      except Exception as e:
```
Bulat-Ziganshin: here

datastore/sqlite.nim
```diff
@@ -0,0 +1,238 @@
...
+    sqlite3_bind_double(s, n.cint, val)
+  elif val is string:
+    try:
+      # `-1` implies string length is num bytes up to first null-terminator;
+      # `SQLITE_TRANSIENT` "indicate[s] that the object is to be copied prior
+      # to the return from sqlite3_bind_*(). The object and pointer to it must
+      # remain valid until then. SQLite will then manage the lifetime of its
+      # private copy."
+      sqlite3_bind_text(s, n.cint, val.cstring, -1.cint, SQLITE_TRANSIENT)
+    except Exception as e:
```
Bulat-Ziganshin: and here: I wonder why you check for exception? it's a C function, like every other sqlite3_* called by bindParam and it can't raise exceptions

datastore/sqlite.nim
```diff
@@ -0,0 +1,238 @@
...
+        sqlite3_bind_blob(s, n.cint, unsafeAddr val[0], val.len.cint,
+          SQLITE_TRANSIENT)
+      except Exception as e:
+        raise (ref Defect)(msg: $e.msg)
+    else:
+      sqlite3_bind_blob(s, n.cint, nil, 0.cint, nil)
+  elif val is int32:
+    sqlite3_bind_int(s, n.cint, val)
+  elif val is uint32:
+    sqlite3_bind_int(s, int(n).cint, int(val).cint)
```
Bulat-Ziganshin: probably it should be `sqlite3_bind_int64(s, n.cint, val)` - definitely you can't pass **uint32** to sqlite3_bind_int accepting **signed** 32-bit int

datastore/sqlite_datastore.nim
```diff
@@ -0,0 +1,272 @@
...
+  $sqlite3_column_text(s, index.cint).cstring
+
+proc dataCol*(
+  s: RawStmtPtr,
+  index = 1): seq[byte] =
+
+  let
+    i = index.cint
+    dataBytes = cast[ptr UncheckedArray[byte]](sqlite3_column_blob(s, i))
+    dataLen = sqlite3_column_bytes(s, i)
```
Bulat-Ziganshin, line 195: In rare cases, sqlite3_column_text and sqlite3_column_bytes may return NULL:
1) there isn't enough memory
2) BLOB is empty
(details are in sqlite docs)

I propose to make an issue for that rather than fixing immediately, since it's neither trivial nor need urgent help

datastore/sqlite_datastore.nim
```diff
@@ -0,0 +1,272 @@
...
+  var
+    exists = false
+
+  proc onData(s: RawStmtPtr) {.closure.} =
+    let
+      v = sqlite3_column_int64(s, 0)
+
+    if v == 1: exists = true
+
+  discard ? self.containsStmt.query((key.id), onData)
```
Bulat-Ziganshin, line 215: I wondered how it works at all: "EXISTS" returns boolean value that's represented by integer: https://www.sqlite.org/datatype3.html#boolean_datatype . Please add that as comment

datastore/sqlite_datastore.nim
```diff
@@ -0,0 +1,272 @@
...
+
+  success exists
```
Bulat-Ziganshin, line 217: I don't know Nim enough - if containsStmt.query returns an error message, will `discard` ignore it or return immediately instead of going to the `success exists` statement?

datastore/sqlite.nim
```diff
@@ -0,0 +1,238 @@
...
+      v = sqlite3_step(s)
+
+    case v
+    of SQLITE_ROW:
+      onData(s)
+      res = success true
+    of SQLITE_DONE:
+      break
+    else:
+      res = failure $sqlite3_errstr(v)
```
Bulat-Ziganshin, line 201: I think we should do `break` here, both for speed and to prevent overwriting error message on the next iteration

datastore/sqlite.nim
```diff
@@ -0,0 +1,238 @@
...
+  checkExec(s)
+
+template dispose*(db: SQLite) =
+  discard sqlite3_close(db)
+
+template dispose*(rawStmt: RawStmtPtr) =
+  discard sqlite3_finalize(rawStmt)
+
+template dispose*(sqliteStmt: SQLiteStmt) =
+  discard sqlite3_finalize(RawStmtPtr(sqliteStmt))
```
Bulat-Ziganshin, line 101: I think it may be combined with previous template into

template dispose*(sqliteStmt: SQLiteStmt | RawStmtPtr) =
  discard sqlite3_finalize(RawStmtPtr(sqliteStmt))

datastore/sqlite.nim
```diff
@@ -0,0 +1,238 @@
...
+template open*(
+  dbPath: string,
+  env: var SQLite,
+  flags = 0) =
+
+  checkErr sqlite3_open_v2(dbPath.cstring, addr env, flags.cint, nil)
+
+proc prepare*[Params, Res](
+  T: type SQLiteStmt[Params, Res],
+  env: SQLite,
```
Bulat-Ziganshin, line 156: I think it can be combined with the `prepare` template using combined type:

 T: type (SQLiteStmt[Params, Res] | RawStmtPtr)

?

Bulat-Ziganshin (commented):
I submit review now so you can see my remarks. Will continue to look into code

datastore/sqlite.nim
dryajov: Current versions of nim allow you to handle system level exceptions (unfortunately), `Exception` is the base class of all fatal and non-fatal exceptions. 

Annotating the module/procedure with `raises: Defect` (enabled through `push: {.upraises: [].}` here) will signal the compiler to allow propagating only `Defect`. 

Most likely, something in the guts of sqlite raises an `Exception` which gets flagged by the `upraises/raises` mechanism and subsequently fails to compile. The only way to guarantee safety in such situations is to remap `Exception` into `Defect` and propagate, which should eventually "panic" the process thus avoiding it to continue execution in a potentially inconsistent state.

Bulat-Ziganshin: Probably the difference is callback passed as the last parameter to sqlite3_bind_text/sqlite3_bind_blob. My bad

dryajov: Now, I'm not sure if this is actually the case @michaelsbradleyjr, if the `C` function is not raising anything, there is no reason to do the wrapping.

datastore/sqlite.nim
dryajov: yep, correct it should use the 64 bit version.

datastore/sqlite_datastore.nim
dryajov: It should be an easy fix? Maybe it's worth addressing ASAP to avoid tripping on sporadic failures?

datastore/sqlite_datastore.nim
```diff
@@ -0,0 +1,272 @@
...
+  key: Key): ?!bool {.locks: "unknown".} =
+
+  var
+    exists = false
+
+  proc onData(s: RawStmtPtr) {.closure.} =
+    let
+      v = sqlite3_column_int64(s, 0)
+
+    if v == 1: exists = true
```
dryajov: should be able to simple covert to boolean with `bool(sqlite3_column_int64(s, 0))`, no need for intermediate `v` here.

datastore/sqlite_datastore.nim
dryajov: Nim doesn't allow dangling return values, meaning every return from a proc/func/method call should be assigned to a variable, otherwise it has to be explicitly discarded. In this case `?` unpacks the value stored in the result and returns that, `discard` is then discarding the value unpacked by `?`.

dryajov (approved):
We can merge this for the most part and work on outstanding issues in separate PRs. In addition to what has already been brought up, we're missing:

- an async api
- templated/generic api
- methods/procs returning `seq[byte]` or other large lists of objects should instead take a `var` parameter to avoid copying data around. Unless we want to start using `sink/lent` and view type parameters anticipating the switch to 1.6

datastore/sqlite.nim
michaelsbradleyjr: The need for `try/except` here was a bit of a mystery. I introduced it only because, when passing `SQLITE_TRANSIENT` [instead of `nil`](https://github.com/status-im/nwaku/blob/master/waku/v2/node/storage/sqlite.nim#L133), the compiler failed with error message: `can raise an unlisted exception: Exception`. Annotating the `dispose` callback with `raises: [...]` did not fix the problem.

It would be nice to have an explanation/understanding of what was happening, but with changes to [nim-sqlite3-abi](https://github.com/arnetheduck/nim-sqlite3-abi) that landed earlier today ([fda455cfea2df707dde052034411ce63de218453](https://github.com/arnetheduck/nim-sqlite3-abi/commit/fda455cfea2df707dde052034411ce63de218453)) it's no longer necessary to define `proc sqlite3_bind_text` in `datastore/sqlite.nim` and it's no longer necessary to wrap with `try/except` when passing `SQLITE_TRANSIENT`.

I've pushed another commit to this PR branch accordingly.

datastore/sqlite.nim
michaelsbradleyjr: This was copy-pasted from [nwaku's sqlite.nim](https://github.com/status-im/nwaku/blob/master/waku/v2/node/storage/sqlite.nim#L130-L148). That doesn't make it wrong or right; however, to my understanding:
* `int(val)` here will convert the unsigned `uint32` to Nim's [default integer type](https://nim-lang.org/docs/system.html#int)
* `int(val).cint` thus converts from `uint32` to [*the type `int` in C*](https://nim-lang.org/docs/system.html#cint)

datastore/sqlite_datastore.nim
michaelsbradleyjr: I'll add a comment 👍

datastore/sqlite_datastore.nim
michaelsbradleyjr: Adding to @dryajov's explanation, [`?`](https://github.com/status-im/nim-stew/blob/master/stew/results.nim#L915-L937) here will do an early return in the case `self.containsStmt.query((key.id), onData)` returns an error, i.e. the error will be propagated and L217 (`success exists`) will not be reached.

datastore/sqlite_datastore.nim
michaelsbradleyjr: Good catch, will give it a shot.

datastore/sqlite.nim
dryajov: I suspect that issues might arise in the way that sqlite represents this values internally. If for some reason sqlite decides to pack values ignoring the sign bit, things can get messy. 

I've found at least one example where uint is stored using the 64 bit version - https://cpp.hotexamples.com/site/file?hash=0xd834bc2db84f0c7ee36fd30ecc9c3625e7c0bcd6d32f8fb48ed8c49d5b05a012&fullName=QGIS-master/src/providers/spatialite/qsql_spatialite.cpp&project=Br1ndavoine/QGIS.

datastore/sqlite.nim
dryajov: Yep, that makes sense, probably missing `Defect` annotations or some such.

datastore/sqlite.nim
Bulat-Ziganshin: No problem, I made pull request there too :) https://github.com/status-im/nwaku/pull/1014

1. n is already int, don't mix it with val !
2. when `val >= 1<<31`, using sqlite3_bind_int should store it as negative value. C API accepts signed value, it has no way to know that you had an unsigned one. So, for `values >= 1<<31`, you have to pass them as 64-bit ones (ideally, we should also make a test passing `uint32(1) << 31` and then requesting it as 64-bit value and checking that it's indeed `1u64 << 31`)

Although, honestly speaking, we don't use uint32 values in nimstore-sqlite backend. They, may be, too :)

datastore/sqlite.nim
Bulat-Ziganshin: I bet that without an annotation, `SQLITE_TRANSIENT = cast[callback](-1)` was considered as real callback that can raise any exception

datastore/sqlite_datastore.nim
Bulat-Ziganshin: Nitpicking: sqlite here returns its own representation of bool, which happens to be the same as Nim one. I think it's ok to use just bool(...) since neither one is likely to change, but I find it important to document that - so that reader understands that sqlite bool is returned by EXISTS and converted to Nim bool here

datastore/sqlite.nim
michaelsbradleyjr: I may misunderstand something here and am always happy to learn and be better informed. 😄 

The opening context is, per the commit I pushed recently, [L32](https://github.com/status-im/nim-datastore/blob/initial_impl/datastore/sqlite.nim#L32): `when val is openarray[byte]|seq[byte]: ...`. To my understanding of the Nim manual, [when](https://nim-lang.org/docs/manual.html#statements-and-expressions-when-statement) is evaluated at compile time.

So, when we consider [L44-L45](https://github.com/status-im/nim-datastore/blob/initial_impl/datastore/sqlite.nim#L44-L45) (again, line numbers are per my latest commit), there's generated code that's guaranteed to be working with a `val` that's a `uint32`.

Therefore, per `int(val).cint` it seems we'd always be passing to SQLite an instance of the largest signed integer type supported by the compiler.

SQLite's docs say:
> The INTEGER storage class, for example, includes 7 different integer datatypes of different lengths. [This makes a difference on disk.](https://www.sqlite.org/fileformat2.html#record_format) But as soon as INTEGER values are read off of disk and into memory for processing, they are converted to the most general datatype (8-byte signed integer). And so for the most part, "storage class" is indistinguishable from "datatype" and the two terms can be used interchangeably.

Am I missing anything?

Bulat-Ziganshin:
Now I think that
1. nimstore-sqlite is our best candidate for production-grade backend, at least for codex V1
2. we don't need to make it async, waiting for processing of get/put requests is ok for codex V1 (see discussion in https://github.com/status-im/nim-codex/issues/117 )

Bulat-Ziganshin:
So, what we need to make it production-grade backend? the plan I propose is centered around Codex V1 needs:

1. provide API compatible with Codex, in particular make get/put/... {.async.} 
2. add query call (listing all blocks in the store)
3. implement performance test
4. (may be) implement GC_ref+SQLITE_STATIC+GC_unref optimization (details in https://forum.nim-lang.org/t/9247 )

Citing @michaelsbradleyjr :
> the default page size in SQLite is currently 4096, do you think we should bump it (or provide an option to change it in nim-datastore) given the 8 KB block size?
> https://www.sqlite.org/pragma.html#pragma_page_size
> https://www.sqlite.org/intern-v-extern-blob.html

I propose to make an option in SQLiteDatastore.new to experiment with it

datastore/sqlite_datastore.nim
michaelsbradleyjr: I'll add a comment noting the conversion.

michaelsbradleyjr:
>  propose to make an option in SQLiteDatastore.new to experiment with it

Will add it shortly.

datastore/sqlite_datastore.nim
Bulat-Ziganshin: Well, it will require tests too (both with empty BLOB and with limited memory if it's possible). So, either fix it in this PR or immediately as the second one.

Bulat-Ziganshin:
About my comments regarding "template dispose" and so on: I don't like the programming style of this module that misuses templates to hide `return` statements. And especially I don't like templates like `open` exported into another module and hiding `return` too.

But since it's just a blind copy of waku module, it may be not our problem. We should bugfix it and backport fixes, but fighting with its style isn't worth it.

datastore/sqlite.nim
Bulat-Ziganshin: BTW, you can always switch to permanent link by pressing 'y` in GitHub page - just open https://github.com/status-im/nim-datastore/blob/initial_impl/datastore/sqlite.nim#L44-L45 and press 'y' while being inside page. Now look at new url

michaelsbradleyjr: > n is already int, don't mix it with val !

Sorry, just appreciated that point. Basically everywhere above I referenced `n` I meant to reference `val`. I updated my comments to that effect.

I don't think it changes the substance of what I wrote above, but `int(n).cint` on L45 should definitely be `n.cint`.

Bulat-Ziganshin: there are two variables here - n is index and val is value
1. for index, use just n.cint like in any other line. it's still an int variable
2. now about val:

cint is C int type which is int32 on most modern platforms, even 64-bit ones and defined as int32 in Nim: https://nim-lang.org/docs/system.html#cint

For `uint32 val`, int(val) should be ok since Nim int is 64-bit (on 64-bit platforms). But then `int(val).cint` will convert it into 32-bit integer. Even without that, sqlite3_bind_int accepts C int, i.e. 32-bit signed integer as its last argument and there is no way it can learn that your original value was unsigned value >= 1<<31 rather than signed value represneted with the same 32 bits.

So, the only way to pass full range of uint32 values is to convert them into int64, and I hope that call to sqlite3_bind_int64 will do that (since its last argument defined as int64). Not 100% sure though since I'm new to Nim.

michaelsbradleyjr:
> About my comments regarding "template dispose" and so on: I don't like the programming style of this module that misuses templates to hide `return` statements. And especially I don't like templates like `open` exported into another module and hiding `return` too.

Indeed, not great, though a `template` can be a nice way to avoid a function call. I'll give it some more thought.

> But since it's just a blind copy of waku module, it may be not our problem. We should bugfix it and backport fixes, but fighting with its style isn't worth it.

For what it's worth, and no offense taken by the way, it wasn't a blind copy (cf. commit history):
* I initially copied it into `datastore/sqlite_datastore.nim`
* Worked through (re: the copy-paste) what I experienced as a big tangle and attempted to clean it up to some degree, while realizing the "tangle" is a bit inherent to the double-wrapper (nim-sqlite3-abi + waku/datastore wrapper)
* Finally extracted it into a separate module `datastore/sqlite.nim`

I will note that I'm not very happy re: the current shape of `datastore/sqlite.nim`, but trying to iron out the rough spots seems/ed like it could turn into a rabbit hole, so I got it to where it is currently with hopes of making it better in the future.

datastore/sqlite.nim
michaelsbradleyjr: >  But `then int(val).cint` will convert it into 32-bit integer

According to my understanding of the Nim manual, per compile-time `when ...: elif val is uint32:`, we're already dealing with a 32-bit unsigned integer as the input at runtime, never any other type, so I don't understand how information could be lost.

Thanks for helping me to understand it better 😅

datastore/sqlite_datastore.nim
dryajov: A subsequent PR should be fine, but this would be the first thing I'd fix, just because it's easy to forget and run into spurious failures later on.

datastore/sqlite.nim
dryajov: Looking at this again, I think the discussion seems to revolve around two aspects. The `n` conversion and the `sqlite3_bind_int` call, so let me try to summarize this -

- `n` is a nim `int`, which will be the largest int for the current architecture, on most 64 bit architectures it's 64 bit. Converting directly to `cint` _might_ lead to some information loss because `cuint` is 32 bits, but I think we'll run into sqlite specific limits before that happens, never the less, it might be prudent to declare `n` as `int32`.
- `sqlite3_bind_int` is a bit more confusing, because it isn't clear what the actual expected data type size is - is it 32 bit as per C ints, or is it the largest machine int (nim style). Since we're dealing with C apis, I would assume that this is a 32 bit int. Next, since we're explicitly matching for a `uint32` here with the `when` statement, it should be safe to convert the `uint32` to an `int32`, since they are both 4 bytes wide.

Sqlite seems to store the data on disk according to it's type, but all manipulations happen on 8 byte signed ints:

from https://www.sqlite.org/datatype3.html
```
The INTEGER storage class, for example, includes 7 different integer datatypes of different lengths. [This makes a difference on disk.](https://www.sqlite.org/fileformat2.html#record_format) But as soon as INTEGER values are read off of disk and into memory for processing, they are converted to the most general datatype (8-byte signed integer). And so for the most part, "storage class" is indistinguishable from "datatype" and the two terms can be used interchangeably.
```

datastore/sqlite.nim
dryajov: Yep, this is missing a `break`

michaelsbradleyjr: Ah, good catch, `break` definitely needed. I missed that when converting from early return to assignment in the `else` branch.

datastore/sqlite.nim
michaelsbradleyjr: `n` represents the zero-indexed column number so practically it's fine to do `n.cint`. If we want to be more precise we could specify `n: Natural` or a range that tops out at `hight(int32)`. However, neither `template bindParams` nor `proc bindParam` are exported, and the way they're currently used (with index numbers derived from tuples), I think it's unlikely we'll encounter a problem.

I'm not sure why the nwaku folks hadn't updated `int(n).cint` in the `elif val is uint32` branch of the `when` statement, but I already changed it yesterday since there's no reason it can't be `n.cint` like the others, i.e. unless we want/need to be more precise. It was probably just overlooked at some point.

What we really care about is `val`.

Consider this:
```nim
echo low(uint32)
echo high(uint32)
echo cast[int32](low(uint32))
echo cast[int32](high(uint32))
echo cast[uint32](cast[int32](low(uint32)))
echo cast[uint32](cast[int32](high(uint32)))

echo ""

echo low(int32)
echo high(int32)
echo cast[uint32](low(int32))
echo cast[uint32](high(int32))
echo cast[int32](cast[uint32](low(int32)))
echo cast[int32](cast[uint32](high(int32)))
```
prints:
```
0
4294967295
0
-1
0
4294967295

-2147483648
2147483647
2147483648
2147483647
-2147483648
2147483647
```

So as you say, *"since we're explicitly matching for a `uint32` here with the when statement, it should be safe to convert the `uint32` to an `int32`, since they are both 4 bytes wide."* That's exactly what I had in mind yesterday.

I'm happy to make the suggested change from `sqlite3_bind_int(s, n.cint, int(val).cint)` to `sqlite3_bind_int64(s, n.cint, val.int64)` if it's for the best. I just wanted to understand under what conditions we might run into a problem without the change, and I'm still not convinced we would.

datastore/sqlite_datastore.nim
michaelsbradleyjr: Tracking in #2.

datastore/sqlite_datastore.nim
michaelsbradleyjr: Done.

datastore/sqlite.nim
michaelsbradleyjr: Done.

datastore/sqlite.nim
michaelsbradleyjr: Tracking in #3.

datastore/sqlite.nim
michaelsbradleyjr: Tracking in #3.

datastore/sqlite_datastore.nim
michaelsbradleyjr: Done.

datastore/sqlite.nim
dryajov: I think the break is meant after the `failure` statement?

michaelsbradleyjr:
Commits squashed. Before that I made some quick changes based on the PR comments, and for some changes that will be more involved I created issues.

After tests all pass again, I'll merge it and then create some issues to track what's next as outlined in comments above, e.g. async, page size adjustment, etc.

datastore/sqlite.nim
michaelsbradleyjr: I'm not sure what you mean; I think maybe I inadvertently hadn't pushed by the time you replied. Per the change I made when I wrote *Done*, the code is:
```nim
case v
of SQLITE_ROW:
  onData(s)
  res = success true
of SQLITE_DONE:
  break
else:
  res = failure $sqlite3_errstr(v)
  break
```

PR #14, opened 2022-06-29
impl and tests for AsyncSQLiteDatastore

### LINKED ISSUE #7: Provide `{.async.}` API for `type SQLiteDatastore` (or extension of that type) compatible with e.g. Codex

michaelsbradleyjr:
Closes #7

### DIFF
--- a/tests/datastore/test_async_sqlite_datastore.nim
+++ b/tests/datastore/test_async_sqlite_datastore.nim
@@ -58,8 +58,7 @@ suite "AsyncSQLiteDatastore":
 
     ds.close
 
-    check:
-      ds.store.isNil
+    check: ds.store.isNil
 
   test "put":
     ds = AsyncSQLiteDatastore.new(basePathAbs, filename).get
--- a/tests/datastore/test_async_sqlite_datastore.nim
+++ b/tests/datastore/test_async_sqlite_datastore.nim
@@ -1,0 +1,142 @@
+import std/options
+import std/os
+
+import asynctest/unittest2
+import pkg/stew/results
+
+import ../../datastore/async_sqlite_datastore
+
+const
+  bytes = @[1.byte, 2.byte, 3.byte]
+  key = Key.init("a:b/c/d:e").get
+
+var
+  ds {.threadvar.}: AsyncSQLiteDatastore
+  basePath {.threadvar.}: string
+  basePathAbs {.threadvar.}: string
+  filename {.threadvar.}: string
+  dbPathAbs {.threadvar.}: string
+
+suite "AsyncSQLiteDatastore":
+  setup:
+    # assumes tests/test_all is run from project root, e.g. with `nimble test`
+    basePath = "tests" / "test_data"
+    basePathAbs = getCurrentDir() / basePath
+    filename = "test_store" & dbExt
+    dbPathAbs = basePathAbs / filename
+
+    removeDir(basePathAbs)
+    require(not dirExists(basePathAbs))
+    discard dbPathAbs # suppresses "declared but not used" re: dbPathAbs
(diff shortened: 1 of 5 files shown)


tests/datastore/test_async_sqlite_datastore.nim
```diff
@@ -0,0 +1,142 @@
...
+const
+  bytes = @[1.byte, 2.byte, 3.byte]
+  key = Key.init("a:b/c/d:e").get
+
+var
+  ds {.threadvar.}: AsyncSQLiteDatastore
+  basePath {.threadvar.}: string
+  basePathAbs {.threadvar.}: string
+  filename {.threadvar.}: string
+  dbPathAbs {.threadvar.}: string
```
michaelsbradleyjr, line 18: When using `asynctest/unittest2` I wasn't able to include these in the `setup:` block like I can when using `unittest2`, I'm not sure why.

Also, to avoid a compiler error re: gc safety, I "cheated" by marking them as threadvars; seemed like an okay workaround for the purposes of a test suite.

Bulat-Ziganshin (commented):
I wonder why tests sometimes uses `check` and sometimes `assert`. If there is no difference, I propose to use `check` everywhere

michaelsbradleyjr:
> I wonder why tests sometimes uses `check` and sometimes `assert`. If there is no difference, I propose to use `check` everywhere

Great question. I've been using `assert` in the test suites to create invariants, i.e. assertions that should always hold true independently of the expressions tested in the body of `check: ...`.

If an `assert [expr]` does not hold true, the test will fail outright at that point in the test. If an expression in the body of a `check: ...` does not evaluate to `true`, later expressions in that body will still be checked.

Put another way, if an `assert` fails as I've intended to use `assert`, there is a problem outside the scope (I don't mean lexical) of what's being tested, and therefore other tests need to be revisited.

It may be that I've not used `assert` vs. `check` as consistently as I intended, so there's likely room for improvement. Or we can scrap the idea of using `assert` to create invariants if it doesn't seem helpful enough to be justified.

Bulat-Ziganshin:
Re "assert vs check" - that's great. Besides practical advantages, it translates an idea of "axiom vs contract" directly into the code

tests/datastore/test_async_sqlite_datastore.nim
markspanbroek: asynctest puts the body of `suite` into a proc, so you don't need the `{.threadvar.}` because the variables won't be global.

arnetheduck: if you want isolation between tests, variables should generally go inside a `setup`, else the instance is reused between tests - websock tests suffered from this, where an instance created in one test was used for the next test - it's generally good practice to not have dependencies like this (where it led to an actual issue of double close) - see https://github.com/status-im/nim-websock/pull/115

arnetheduck: also, most nim projects we have prefer the name `asyncTest` for async tests so as to highlight the additional failure modes of running with potential execution reordering

michaelsbradleyjr:
Converted to draft PR because it will likely be closed in favor of #15.

tests/datastore/test_async_sqlite_datastore.nim
Bulat-Ziganshin: @arnetheduck from where to import asyncTest ? 

I find one in https://github.com/status-im/nim-chronos/blob/master/chronos/unittest2/asynctests.nim
 `-` is it proper place?

Bulat-Ziganshin:
Closed beacuse it definitely not what we want

PR #15, opened 2022-07-04
refactor base API to be async

### LINKED ISSUE #7: Provide `{.async.}` API for `type SQLiteDatastore` (or extension of that type) compatible with e.g. Codex

michaelsbradleyjr:
Refactor `Datastore` impls and tests accordingly.

Closes #7.

### DIFF
--- a/datastore/sqlite_datastore.nim
+++ b/datastore/sqlite_datastore.nim
@@ -1,10 +1,12 @@
 import std/os
 import std/times
 
+import pkg/chronos
 import pkg/questionable
 import pkg/questionable/results
 import pkg/sqlite3_abi
 import pkg/stew/byteutils
+from pkg/stew/results as stewResults import get, isErr
 import pkg/upraises
 
 import ./datastore
@@ -204,7 +206,7 @@ proc timestampCol*(
 
 method contains*(
   self: SQLiteDatastore,
-  key: Key): ?!bool {.locks: "unknown".} =
+  key: Key): Future[?!bool] {.async, locks: "unknown".} =
 
   var
     exists = false
@@ -212,22 +214,25 @@ method contains*(
   proc onData(s: RawStmtPtr) {.closure.} =
     exists = sqlite3_column_int64(s, 0).bool
 
-  discard ? self.containsStmt.query((key.id), onData)
+  let
+    queryRes = self.containsStmt.query((key.id), onData)
 
-  success exists
+  if queryRes.isErr: return failure queryRes.error.msg
+
+  return success exists
 
 method delete*(
   self: SQLiteDatastore,
-  key: Key): ?!void {.locks: "unknown".} =
+  key: Key): Future[?!void] {.async, locks: "unknown".} =
 
   if self.readOnly:
-    failure "database is read-only":
+    return failure "database is read-only":
   else:
-    self.deleteStmt.exec((key.id))
+    return self.deleteStmt.exec((key.id))
 
 method get*(
   self: SQLiteDatastore,
-  key: Key): ?!(?seq[byte]) {.locks: "unknown".} =
+  key: Key): Future[?!(?seq[byte])] {.async, locks: "unknown".} =
 
   # see comment in ./filesystem_datastore re: finer control of memory
   # allocation in `method get`, could apply here as well if bytes were read
@@ -240,33 +245,35 @@ method get*(
     bytes = dataCol(s, 0)
 
   let
-    exists = ? self.getStmt.query((key.id), onData)
+    queryRes = self.getStmt.query((key.id), onData)
 
-  if exists:
-    success bytes.some
+  if queryRes.isErr: return failure queryRes.error.msg
+
+  if queryRes.get:
+    return success bytes.some
   else:
-    success seq[byte].none
+    return success seq[byte].none
 
 proc put*(
   self: SQLiteDatastore,
   key: Key,
-  data: openArray[byte],
-  timestamp: int64): ?!void =
+  data: seq[byte],
+  timestamp: int64): Future[?!void] {.async.} =
 
   if self.readOnly:
-    failure "database is read-only"
+    return failure "database is read-only"
   else:
-    self.putStmt.exec((key.id, @data, timestamp))
+    return self.putStmt.exec((key.id, @data, timestamp))
 
 method put*(
   self: SQLiteDatastore,
   key: Key,
-  data: openArray[byte]): ?!void {.locks: "unknown".} =
+  data: seq[byte]): Future[?!void] {.async, locks: "unknown".} =
 
-  self.put(key, data, timestamp())
+  return await self.put(key, data, timestamp())
 
 # method query*(
 #   self: SQLiteDatastore,
-#   query: ...): ?!(?...) {.locks: "unknown".} =
+#   query: ...): Future[?!(?...)] {.async, locks: "unknown".} =
 #
-#   success ....none
+#   return success ....none
--- a/datastore/sqlite_datastore.nim
+++ b/datastore/sqlite_datastore.nim
@@ -217,7 +217,7 @@ method contains*(
   let
     queryRes = self.containsStmt.query((key.id), onData)
 
-  if queryRes.isErr: return failure queryRes.error.msg
+  if queryRes.isErr: return queryRes
 
   return success exists
 
@@ -274,4 +274,4 @@ method put*(
 #   self: SQLiteDatastore,
 #   query: ...): Future[?!(?...)] {.async, locks: "unknown".} =
 #
-#   return success ....none
+#   return success ....some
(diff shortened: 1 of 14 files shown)


datastore/sqlite_datastore.nim
```diff
@@ -240,33 +245,35 @@ method get*(
     bytes = dataCol(s, 0)
 
   let
-    exists = ? self.getStmt.query((key.id), onData)
+    queryRes = self.getStmt.query((key.id), onData)
+
+  if queryRes.isErr: return failure queryRes.error.msg
```
Bulat-Ziganshin: why not just `return queryRes` - here and in the lot of similar places?

michaelsbradleyjr: I may be able to do that in some places, but in others it wasn't working. I'll check again.

tests/datastore/test_filesystem_datastore.nim
```diff
@@ -65,37 +67,37 @@ suite "FileSystemDatastore":
...
 
-    putRes = ds.put(key, bytes)
+    putRes = await ds.put(key, bytes)
 
     check:
       putRes.isOk
       readFile(path).toBytes == bytes
 
-  test "delete":
+  asyncTest "delete":
```
Bulat-Ziganshin, line 100: I wonder why you are using asyncTest==test. In Codex, all tests are using plain `test`

datastore/sqlite_datastore.nim
michaelsbradleyjr: Ah, of course, it's because of the type mismatch. The return type (wrapped in `Future`) for `method get` is `?!(?seq[byte])` while for `proc query` it's `?!bool`.

If we do `if queryRes.isErr: return queryRes` then the compiler fails with:
```
Error: type mismatch: got <Future[Result[Option[seq[byte]], ref CatchableError]], Result[system.bool, ref CatchableError]> ... but expression 'queryRes' is of type: Result[system.bool, ref CatchableError]
```

tests/datastore/test_filesystem_datastore.nim
michaelsbradleyjr: See this comment: https://github.com/status-im/nim-datastore/pull/14#discussion_r911044894.

`test` coming from `pkg/asynctest` is https://github.com/status-im/asynctest/blob/main/asynctest/templates.nim#L34-L37

datastore/sqlite_datastore.nim
```diff
@@ -240,33 +245,35 @@ method get*(
...
 
-  if exists:
-    success bytes.some
+  if queryRes.get:
+    return success bytes.some
   else:
-    success seq[byte].none
+    return success seq[byte].none
```
Bulat-Ziganshin: what about 
```Nim
var bytes: ?seq[byte] = none
...
if queryRes.isErr:
  return queryRes
else:
  return success bytes
```
?

datastore/tiered_datastore.nim
```diff
@@ -26,57 +28,94 @@ proc stores*(self: TieredDatastore): seq[Datastore] =
...
-  success exists
+  return success exists
 
 method delete*(
   self: TieredDatastore,
-  key: Key): ?!void {.locks: "unknown".} =
+  key: Key): Future[?!void] {.async, locks: "unknown".} =
+
+  var
+    pending: seq[Future[?!void]]
```
Bulat-Ziganshin: What about shorter version:
```
let pending = await allFinished(
  self.stores.mapIt( it.delete(key) ))
```
It applies to get, put, delete.

datastore/tiered_datastore.nim
```diff
@@ -26,57 +28,94 @@ proc stores*(self: TieredDatastore): seq[Datastore] =
 
 method contains*(
   self: TieredDatastore,
-  key: Key): ?!bool {.locks: "unknown".} =
+  key: Key): Future[?!bool] {.async, locks: "unknown".} =
 
   var
     exists = false
```
Bulat-Ziganshin: I think it will be shorter and probably cleaner without the `exists` variable

datastore/sqlite_datastore.nim
michaelsbradleyjr: For that one if we do `if queryRes.isErr: return queryRes` then the compiler fails with:
```
Error: type mismatch: got <Future[Result[Option[seq[byte]], ref CatchableError]], Result[system.bool, ref CatchableError]> ... but expression 'queryRes' is of type: Result[system.bool, ref CatchableError]
```

In any case what we want to return is the seq in `var bytes`, which is populated via side-effect in the query callback.

datastore/tiered_datastore.nim
michaelsbradleyjr: I thought about that, but  I was also thinking we want to return `failure` if any of the put/delete ops had an error. It's not clear to me how to do that via `mapIt`.

For `get` we want to step through the stores from first in the list (seq) to last, i.e. we don't want to run the `get` ops concurrently.

datastore/sqlite_datastore.nim
Bulat-Ziganshin: You are right - there are places where failure types are different. But in the places when they match, I propose to return the same failure object rather than making a new one from the failure message

datastore/tiered_datastore.nim
michaelsbradleyjr: The natural alternative would be:
```
...
if containsRes.get: break

return success containsRes.get
```

For that to work I'd need to declare `var containsRes` outside of the scope of the `for` loop, i.e. in the same place `var exists` is currently declared.  Altogether that doesn't seem much cleaner than the approach I took with `var exists` but I can make the change if that's what is preferred.

tests/datastore/test_filesystem_datastore.nim
Bulat-Ziganshin: I'm not sure that he meant to define `asyncTest=test`. Asked him, let's wait for answer

datastore/sqlite_datastore.nim
Bulat-Ziganshin: yes, we already discussed isErr part in another thread.

what about the `else` part - instead of checking queryRes.get, just return `bytes` that is initialized with `none`, but can be overwritten in onData

datastore/tiered_datastore.nim
Bulat-Ziganshin: the full code:
```
  let pending = await allFinished(
    self.stores.mapIt( it.delete(key) ))

  for fut in pending:
    if fut.read().isErr: return fut.read()

  return success()
```

You are right - it applies only to delete and put

datastore/sqlite_datastore.nim
michaelsbradleyjr: Change made: https://github.com/status-im/nim-datastore/pull/15/commits/2a7e1b987e2e8f2687560e06d58ab035432b307f.

Thanks for spotting that!

datastore/tiered_datastore.nim
michaelsbradleyjr: Ah, I see now, I was reading `allFinished` as `allFutures`, the latter returns `Future[void]` but `allFinished` gives us the seq we want, will make the change.

datastore/tiered_datastore.nim
Bulat-Ziganshin: I mean smth like that, the main problem is how to make it clear that we return true on the first containsRes==true:
```
for store in self.stores:
    let containsRes = await store.contains(key)
    if containsRes.isErr: return containsRes
    if containsRes.get == true: return success true

  return success false
```
It's 5 non-empty lines instead of 7

datastore/tiered_datastore.nim
michaelsbradleyjr: Done.

datastore/tiered_datastore.nim
michaelsbradleyjr: Change made.

datastore/sqlite_datastore.nim
michaelsbradleyjr: I reviewed all usage of `return`, `failure`, `success` and made relevant changes.

tests/datastore/test_filesystem_datastore.nim
Bulat-Ziganshin: so it's the last thing to resolve. If you want, you can merge existing code and then fix it (if required) in next PR

michaelsbradleyjr: Let's merge it, and revisit `asyncTest=test` depending on his answer to your question.

michaelsbradleyjr:
Ah, whoops, I meant to squash the `[wip] ...` commits. Doing a quick force-push fixup given that this library currently has no dependents.

Bulat-Ziganshin:
definitely noone will mention :)

nice job. this week we will make the sqlite backend working in Codex

PR #17, opened 2022-07-06
check for out-of-memory errors re: sqlite3_column_text/blob/bytes

### LINKED ISSUE #2: Add explicit handling for cases when sqlite3_column_text and sqlite3_column_bytes may return NULL
See: https://github.com/status-im/nim-datastore/pull/1#discussion_r901448334

michaelsbradleyjr:
Check for out-of-memory errors re: usage of `sqlite3_column_text`, `sqlite3_column_blob`, `sqlite3_column_bytes`; raise a `Defect` if a SQLite out-of-memory error is encountered.

Closes #2.

---

Usage of the functions listed above is related to the helpers `idCol` and `dataCol`. Those procs can be made more resilient by checking column metadata with `sqlite3_column_origin_name()`.  Enabling that function involves passing `-DSQLITE_ENABLE_COLUMN_METADATA` to the C compiler, which is handled automatically in `datastore/sqlite.nim`.

The goal is to detect mismatches between caller-supplied indexes and original column names, and in that case crash the process by raising `Defect`. This should help avoid harder to debug errors for such mismatches, which would trigger automatic SQLite type conversions.

These helper procs (including `timestampCol`) are now higher-order, which allows column metadata checks to be run only once, i.e. when setting up derivative helpers to be used in an `onData` callback.

Related to changes to `idCol`, etc., use compile-time constants for column names and default indexes.

Also adjust callback proc annotations to be more precise, and remove unnecessary annotations.

---

Suppress overly verbose hints and warnings at compile-time by adding some switches to `config.nims`.

### DIFF
--- a/datastore/sqlite.nim
+++ b/datastore/sqlite.nim
@@ -143,25 +143,11 @@ template journalModePragmaStmt*(env: SQLite): RawStmtPtr =
     return failure $sqlite3_errstr(x)
 
   let
-    x = sqlite3_column_text(s, 0).cstring
+    x = $sqlite3_column_text_not_null(s, 0)
 
-  # detect out-of-memory error
-  # see the conversion table and final paragraph of:
-  # https://www.sqlite.org/c3ref/column_blob.html
-  # see also https://www.sqlite.org/rescode.html
-
-  # in order to detect an out-of-memory error check that the result is a null
-  # pointer and that the result code is an error code
-  if x.isNil:
-    let
-      code = sqlite3_errcode(sqlite3_db_handle(s))
-
-    if not (code in [SQLITE_OK, SQLITE_ROW, SQLITE_DONE]):
-      raise (ref Defect)(msg: $sqlite3_errstr(code))
-
-  if not ($x in ["memory", "wal"]):
+  if not (x in ["memory", "wal"]):
     s.dispose
-    return failure "Invalid pragma result: \"" & $x & "\""
+    return failure "Invalid pragma result: \"" & x & "\""
 
   s
 
@@ -245,3 +231,21 @@ proc query*(
 proc release*[T](x: var AutoDisposed[T]): T =
   result = x.val
   x.val = nil
+
+proc sqlite3_column_text_not_null*(
+  s: RawStmtPtr,
+  index: cint): cstring =
+
+  let
+    text = sqlite3_column_text(s, index).cstring
+
+  if text.isNil:
+    # see the conversion table and final paragraph of:
+    # https://www.sqlite.org/c3ref/column_blob.html
+    # a null pointer here implies an out-of-memory error
+    let
+      code = sqlite3_errcode(sqlite3_db_handle(s))
+
+    raise (ref Defect)(msg: $sqlite3_errstr(code))
+
+  text
--- a/datastore/sqlite.nim
+++ b/datastore/sqlite.nim
@@ -3,16 +3,22 @@ import pkg/questionable/results
 import pkg/sqlite3_abi
 import pkg/upraises
 
-push: {.upraises: [].}
-
 # Adapted from:
 # https://github.com/status-im/nwaku/blob/master/waku/v2/node/storage/sqlite.nim
+
+# see https://www.sqlite.org/c3ref/column_database_name.html
+# can pass `--forceBuild:on` to the Nim compiler if a SQLite build without
+# `-DSQLITE_ENABLE_COLUMN_METADATA` option is stuck in the build cache,
+# e.g. `nimble test --forceBuild:on`
+{.passC: "-DSQLITE_ENABLE_COLUMN_METADATA".}
+
+push: {.upraises: [].}
 
 type
   AutoDisposed*[T: ptr|ref] = object
     val*: T
 
-  DataProc* = proc(s: RawStmtPtr) {.closure.}
+  DataProc* = proc(s: RawStmtPtr) {.closure, gcsafe.}
 
   NoParams* = tuple # empty tuple
 
--- a/datastore/sqlite.nim
+++ b/datastore/sqlite.nim
@@ -136,9 +136,26 @@ template journalModePragmaStmt*(env: SQLite): RawStmtPtr =
     s.dispose
     return failure $sqlite3_errstr(x)
 
-  if (let x = $sqlite3_column_text(s, 0).cstring; x != "memory" and x != "wal"):
+  let
+    x = sqlite3_column_text(s, 0).cstring
+
+  # detect out-of-memory error
+  # see the conversion table and final paragraph of:
+  # https://www.sqlite.org/c3ref/column_blob.html
+  # see also https://www.sqlite.org/rescode.html
+
+  # in order to detect an out-of-memory error check that the result is a null
+  # pointer and that the result code is an error code
+  if x.isNil:
+    let
+      code = sqlite3_errcode(sqlite3_db_handle(s))
+
+    if not (code in [SQLITE_OK, SQLITE_ROW, SQLITE_DONE]):
+      raise (ref Defect)(msg: $sqlite3_errstr(code))
+
+  if not ($x in ["memory", "wal"]):
     s.dispose
-    return failure "Invalid pragma result: " & $x
+    return failure "Invalid pragma result: \"" & $x & "\""
 
   s
 
--- a/datastore/sqlite_datastore.nim
+++ b/datastore/sqlite_datastore.nim
@@ -97,24 +97,26 @@ const
     ) VALUES (?, ?, ?);
   """
 
+template checkColMetadata(s: RawStmtPtr, i: int, expectedName: string) =
+  let
+    colName = sqlite3_column_origin_name(s, i.cint)
+
+  if colName.isNil:
+    raise (ref Defect)(msg: "no column exists for index " & $i)
+
+  if $colName != expectedName:
+    raise (ref Defect)(msg: "original column name for index " & $i & " was \"" &
+      $colName & "\" but expected \"" & expectedName & "\"")
+
 proc idCol*(
   s: RawStmtPtr,
   index = idColIndex): BoundIdCol =
 
-  let
-    i = index.cint
-    colName = sqlite3_column_origin_name(s, i)
-
-  if colName.isNil:
-    raise (ref Defect)(msg: "no column exists for index " & $index)
-
-  if $colName != idColName:
-    raise (ref Defect)(msg: "original column name for index " & $index &
-      " was \"" & $colName & "\" but expected \"" & idColName & "\"")
+  checkColMetadata(s, index, idColName)
 
   return proc (): string =
     let
-      text = sqlite3_column_text(s, i)
+      text = sqlite3_column_text(s, index.cint)
 
     # detect out-of-memory error
     # see the conversion table and final paragraph of:
@@ -134,19 +136,11 @@ proc dataCol*(
   s: RawStmtPtr,
   index = dataColIndex): BoundDataCol =
 
-  let
-    i = index.cint
-    colName = sqlite3_column_origin_name(s, i)
-
-  if colName.isNil:
-    raise (ref Defect)(msg: "no column exists for index " & $index)
-
-  if $colName != dataColName:
-    raise (ref Defect)(msg: "original column name for index " & $index &
-      " was \"" & $colName & "\" but expected \"" & dataColName & "\"")
+  checkColMetadata(s, index, dataColName)
 
   return proc (): seq[byte] =
     let
+      i = index.cint
       blob = sqlite3_column_blob(s, i)
 
     # detect out-of-memory error
@@ -183,19 +177,10 @@ proc timestampCol*(
   s: RawStmtPtr,
   index = timestampColIndex): BoundTimestampCol =
 
-  let
-    i = index.cint
-    colName = sqlite3_column_origin_name(s, i)
-
-  if colName.isNil:
-    raise (ref Defect)(msg: "no column exists for index " & $index)
-
-  if $colName != timestampColName:
-    raise (ref Defect)(msg: "original column name for index " & $index &
-      " was \"" & $colName & "\" but expected \"" & timestampColName & "\"")
+  checkColMetadata(s, index, timestampColName)
 
   return proc (): int64 =
-    sqlite3_column_int64(s, i)
+    sqlite3_column_int64(s, index.cint)
 
 proc new*(
   T: type SQLiteDatastore,
--- a/datastore/sqlite_datastore.nim
+++ b/datastore/sqlite_datastore.nim
@@ -160,15 +160,6 @@ proc dataCol*(
 
     let
       dataLen = sqlite3_column_bytes(s, i)
-
-    # an out-of-memory error can be inferred from a null pointer result
-    if (unsafeAddr dataLen).isNil:
-      let
-        code = sqlite3_errcode(sqlite3_db_handle(s))
-
-      raise (ref Defect)(msg: $sqlite3_errstr(code))
-
-    let
       dataBytes = cast[ptr UncheckedArray[byte]](blob)
 
     @(toOpenArray(dataBytes, 0, dataLen - 1))
--- a/datastore/sqlite_datastore.nim
+++ b/datastore/sqlite_datastore.nim
@@ -97,7 +97,7 @@ const
     ) VALUES (?, ?, ?);
   """
 
-template checkColMetadata(s: RawStmtPtr, i: int, expectedName: string) =
+proc checkColMetadata(s: RawStmtPtr, i: int, expectedName: string) =
   let
     colName = sqlite3_column_origin_name(s, i.cint)
 
--- a/datastore/sqlite_datastore.nim
+++ b/datastore/sqlite_datastore.nim
@@ -185,7 +185,22 @@ proc idCol*(
   s: RawStmtPtr,
   index = 0): string =
 
-  $sqlite3_column_text(s, index.cint).cstring
+  let
+    text = sqlite3_column_text(s, index.cint).cstring
+
+  # detect out-of-memory error
+  # see the conversion table and final paragraph of:
+  # https://www.sqlite.org/c3ref/column_blob.html
+
+  # the "id" column is NOT NULL PRIMARY KEY so an out-of-memory error can be
+  # inferred from a null pointer result
+  if text.isNil:
+    let
+      code = sqlite3_errcode(sqlite3_db_handle(s))
+
+    raise (ref Defect)(msg: $sqlite3_errstr(code))
+
+  $text
 
 proc dataCol*(
   s: RawStmtPtr,
@@ -193,8 +208,35 @@ proc dataCol*(
 
   let
     i = index.cint
-    dataBytes = cast[ptr UncheckedArray[byte]](sqlite3_column_blob(s, i))
+    blob = sqlite3_column_blob(s, i)
+
+  # detect out-of-memory error
+  # see the conversion table and final paragraph of:
+  # https://www.sqlite.org/c3ref/column_blob.html
+  # see also https://www.sqlite.org/rescode.html
+
+  # the "data" column can be NULL so in order to detect an out-of-memory error
+  # it is necessary to check that the result is a null pointer and that the
+  # result code is an error code
+  if blob.isNil:
+    let
+      code = sqlite3_errcode(sqlite3_db_handle(s))
+
+    if not (code in [SQLITE_OK, SQLITE_ROW, SQLITE_DONE]):
+      raise (ref Defect)(msg: $sqlite3_errstr(code))
+
+  let
     dataLen = sqlite3_column_bytes(s, i)
+
+  # an out-of-memory error can be inferred from a null pointer result
+  if (unsafeAddr dataLen).isNil:
+    let
+      code = sqlite3_errcode(sqlite3_db_handle(s))
+
+    raise (ref Defect)(msg: $sqlite3_errstr(code))
+
+  let
+    dataBytes = cast[ptr UncheckedArray[byte]](blob)
 
   @(toOpenArray(dataBytes, 0, dataLen - 1))
 
--- a/datastore/sqlite_datastore.nim
+++ b/datastore/sqlite_datastore.nim
@@ -117,22 +117,7 @@ proc idCol*(
   checkColMetadata(s, index, idColName)
 
   return proc (): string =
-    let
-      text = sqlite3_column_text(s, index.cint)
-
-    # detect out-of-memory error
-    # see the conversion table and final paragraph of:
-    # https://www.sqlite.org/c3ref/column_blob.html
-
-    # the "id" column is NOT NULL PRIMARY KEY so an out-of-memory error can be
-    # inferred from a null pointer result
-    if text.isNil:
-      let
-        code = sqlite3_errcode(sqlite3_db_handle(s))
-
-      raise (ref Defect)(msg: $sqlite3_errstr(code))
-
-    $text.cstring
+    $sqlite3_column_text_not_null(s, index.cint)
 
 proc dataCol*(
   s: RawStmtPtr,
--- a/datastore/sqlite_datastore.nim
+++ b/datastore/sqlite_datastore.nim
@@ -52,10 +52,6 @@ const
   dataColName = "data"
   timestampColName = "timestamp"
 
-  idColIndex = 0
-  dataColIndex = 1
-  timestampColIndex = 2
-
   idColType = "TEXT"
   dataColType = "BLOB"
   timestampColType = "INTEGER"
@@ -110,7 +106,7 @@ proc checkColMetadata(s: RawStmtPtr, i: int, expectedName: string) =
 
 proc idCol*(
   s: RawStmtPtr,
-  index = idColIndex): BoundIdCol =
+  index: int): BoundIdCol =
 
   checkColMetadata(s, index, idColName)
 
@@ -134,7 +130,7 @@ proc idCol*(
 
 proc dataCol*(
   s: RawStmtPtr,
-  index = dataColIndex): BoundDataCol =
+  index: int): BoundDataCol =
 
   checkColMetadata(s, index, dataColName)
 
@@ -175,7 +171,7 @@ proc dataCol*(
 
 proc timestampCol*(
   s: RawStmtPtr,
-  index = timestampColIndex): BoundTimestampCol =
+  index: int): BoundTimestampCol =
 
   checkColMetadata(s, index, timestampColName)
 
--- a/datastore/sqlite_datastore.nim
+++ b/datastore/sqlite_datastore.nim
@@ -48,9 +48,9 @@ const
   dbExt* = ".sqlite3"
   tableName* = "Store"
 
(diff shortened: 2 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -133,4 +133,4 @@ jobs:
           source "${HOME}/.bash_env"
           nim --version
           echo
-          nimble --verbose test
+          nimble --verbose test --hints:off
```
Menduist: This also suppresses useful hints (ie XDeclaredButNotUsed)
See https://github.com/status-im/nim-libp2p/blob/master/config.nims

michaelsbradleyjr: Thanks!

Bulat-Ziganshin:
Is it finished and ready for review?

michaelsbradleyjr:
It is finished and I marked it as ready for review.

Bulat-Ziganshin:
>Those procs can be made more resilient by checking column metadata

One part of module spying of other part inside the same module is path to madness

>This should help avoid harder to debug errors for such mismatches

There is simpler way to do that. It's not ideal, but by decreasing complexity substantially, IMHO it's better compromise:

1. get rid of these consts, instead just paste their values into strings:
```
  tableName* = "Store"

  idColName = "id"
  dataColName = "data"
  timestampColName = "timestamp"

  idColIndex = 0
  dataColIndex = 1
  timestampColIndex = 2

  idColType = "TEXT"
  dataColType = "BLOB"
  timestampColType = "INTEGER"
```

but use more unique names for table and columns, f.e.
```
  tableName = "NimDataStoreTable"
  idColName = "ndsId"
  dataColName = "ndsData"
  timestampColName = "ndsTimestamp"
```

2. Since column numbers are SQL_statement-specific, specify them next to corresponding statements:

```
  getStmtStr = """
    SELECT ndsData FROM NimDataStoreTable
    WHERE ndsId = ?;
  """
  getStmtDataColId = 0
```

3. return back old types of idCol/dataCol..., but don't provide default values to avoid inattention bugs:
```
proc idCol*(
  s: RawStmtPtr,
  index: int): string =
```
note "= 0" replaced by "int"

michaelsbradleyjr:
@Bulat-Ziganshin

> One part of module spying of other part inside the same module...

It was not my intention to implement such "spying" behavior, and I do not understand where/how in the changes in the second commit in this PR (2eab6f949b2354dd08a2b2920719ee381b440fbe) you think that's being done. Can you elaborate?

Jumping to your third point:

> 3. return back old types of idCol/dataCol..., but don't provide default values to avoid inattention bugs:
> ```
> proc idCol*(
>   s: RawStmtPtr,
>   index: int): string =
> ```
> note "= 0" replaced by "int"

Let the context be the first commit in this PR's branch (263f4f1097ea4b5b8ea167639be229502769777f) and the test suite `tests/test_sqlite_datastore.nim`.

Assume some data is in the database:
```nim
let
  key = Key.init("a:b/c/d:e").get
  putRes = await ds.put(key, @[1.byte, 2.byte, 3.byte])

assert putRes.isOk
```

Let's try this query:
```nim
let
  query = "SELECT data FROM " & tableTitle & " LIMIT 1;"
```
And this callback:
```nim
var
  qData: seq[byte]

proc onData(s: RawStmtPtr) {.closure.} =
  qData = dataCol(s, 1)
```
With some probable but mistaken reasoning for `dataCol(s, 1)` such as:
> `data` is the second column per `CREATE TABLE ...`, and columns are zero-indexed, so `1` seems like the correct choice for the `index` argument in `dataCol(s, index)`.

When executed with:
```nim
var
  qRes = ds.env.query(query, onData)
```
There will be an error, i.e. `echo qRes.error.msg` gives:
```
Unhandled exception: column index out of range [Defect]
```

For the prior commit on `master` (1a1a1ac64208e4181077c4d6a84248ff7c1e537f) the situation is worse, i.e. `qRes.isOk` is `true` and `qData` is/remains an empty seq `@[]`.

With the changes I made in 2eab6f949b2354dd08a2b2920719ee381b440fbe, and with `tableTitle` changed to `tableName` in the code above, when `qData = dataCol(s, 1)()` is executed a `Defect` will be raised, with message:
```
Unhandled exception: original column name for index 1 was not "data" [Defect]
```

You may already understand what's happening: `index` in `dataCol(s, index)` must match with the column order in `SELECT ...` regardless of the order in `CREATE TABLE ...`.

So we're dealing with something that will manifest caller errors only at runtime, and to my mind it's less a matter of *"inattention"* and more a requirement that's easy to misunderstand.

Stopping it dead in its tracks, with the help of [`sqlite3_column_origin_name()`](https://www.sqlite.org/c3ref/column_database_name.html), seemed to me to fit well with the sentiment [*"it's easy to forget and run into spurious failures later on"*](https://github.com/status-im/nim-datastore/pull/1#discussion_r903090761), which was part of the discussion motivating #2 and this PR. This is purely a facility provided by SQLite at runtime, it has nothing to do with parts of the `datastore/sqlite_datastore.nim` module spying on one another.

As to your first and second points, I don't see how your suggested changes help to alleviate this problem. It's not a matter of naming, it's a matter of the caller supplying an incorrect `index`. Unfortunately, I don't think there's a way to catch such mistakes at compile-time.

If we check column metadata at runtime, then a mistake by the caller is likely to be flagged by runtime crashes early on, with a helpful error message. Otherwise, the mistake will likely manifest later as a runtime crash that is more mysterious.

My reason for defining column names and default indexes as compile-time constants is to make the implementation less error-prone if/when we change it. Rather than having repeated strings/ints spread through the module, they're gathered in one `const ...` block.

In the case where the column index is not out of range but matches the wrong column, I see that we can do better, i.e. the message for the raised `Defect` could be something like:
```
original column name for index 1 was "id" but the specified name was "data"
```
I'll push a change to that effect.

Bulat-Ziganshin:
Yeah, here we have two mutually dependent codes:
1. `let query = "SELECT data FROM " & tableTitle & " LIMIT 1;"`
2. `qData = dataCol(s, 1)`
 
You added 3rd code to protect against inconsistency between 1st code and 2nd code, and that's exactly example of what I called "one part of module spying on other".

Instead I propose to minimize chances that they may become inconsistent by placing 1 & 2 together:

```Nim
let 
  getQuery = "SELECT data FROM NimDataStoreTable LIMIT 1;"
  getStmtDataColId = 0
```
and add corresponding comments if necessary. This way, if someone will ever edit the query, they will edit the id too.

Of course, one may fail even in this case, but checking against every possible mistake is overengineering. We already made it super-easy to keep both in check.


>As to your first and second points, I don't see how your suggested changes help to alleviate this problem.
>My reason for defining column names and default indexes as compile-time constants is to make the implementation less error-prone if/when we change it. 

idea of right architecture is to make it easy to change things that we can probably change, rather than make everything flexible. In the latter case, we overengineer it, adding extra complexity without real benefits.

It's why I propose to drop special constants for the table/field names, and instead choose names that are unique inside the module and paste them directly into SQL statements. This way, we can easily replace them in text editor if ever need, but the statements will not be cluttered with all those concatenations. It just improves readability.

Bulat-Ziganshin:
> Stopping it dead in its tracks, with the help of [`sqlite3_column_origin_name()`](https://www.sqlite.org/c3ref/column_database_name.html), seemed to me to fit well with the sentiment [*"it's easy to forget and run into spurious failures later on"*](https://github.com/status-im/nim-datastore/pull/1#discussion_r903090761)

There are external problems which we can't avoid - memory overflow, corrupted files, incorrect user input. We have no choice but to check all these conditions meticulously. In particular, empty BLOB or memory shortage are examples of these external problems.

OTOH, some modules (or packages we have imported) may contain bugs. It's ok to habe some checks, especially against external packages. But in this particular case you added check against other part of the same module. Moreover, it's rather comlicated check, adding to overall module complexity. Since complexity is source of errors, it's fighting fire with gasoline.

Moreover, there is a simple way to make such errors unlikely - put SQL statement and its column index together.

>For the prior commit on master (https://github.com/status-im/nim-datastore/commit/1a1a1ac64208e4181077c4d6a84248ff7c1e537f) the situation is worse, i.e. qRes.isOk is true and qData is/remains an empty seq @[].

let's check that qRes.isOk AND qData isn't empty. It's short, fast and makes double-check just for case.

Bulat-Ziganshin:
BTW, I think we should add "NOT NULL" specifier for the BLOB field too.

michaelsbradleyjr:
> There are external problems which we can't avoid - memory overflow, corrupted files, incorrect user input. We have no choice but to check all these conditions meticulously. In particular, empty BLOB or memory shortage are examples of these external problems.
> 
> OTOH, some modules (or packages we have imported) may contain bugs. It's ok to habe some checks, especially against external packages. But in this particular case you added check against other part of the same module. Moreover, it's rather comlicated check, adding to overall module complexity. Since complexity is source of errors, it's fighting fire with gasoline.

I re-read the latest replies, and I realized that part of the difficulty is that we have different scopes in mind.

You're viewing the changes I introduced in 2eab6f949b2354dd08a2b2920719ee381b440fbe with respect to the organization and logic of the `datastore/sqlite_datastore.nim` module itself.

I was thinking of them, almost entirely, in terms of external usage of `idCol`, `dataCol`, `timestampCol`, i.e. by a user of the library (could be ourselves in a Codex module, but also anyone else in their own projects) who for whatever reason wants to write a custom query and callback for `ds.env`.

If we're not attempting to address external usage (imagine we could even go so far as not exporting `idCol`, etc.) then indeed the changes in 2eab6f949b2354dd08a2b2920719ee381b440fbe are pointless, and I would not have proposed them in the first place. My goal was to help a caller external to the module avoid an easy kind of mistake by detecting it early on and giving a helpful error message.

How about I drop 2eab6f949b2354dd08a2b2920719ee381b440fbe from this PR and then review can focus on the changes in 263f4f1097ea4b5b8ea167639be229502769777f?

I can open another PR with 2eab6f949b2354dd08a2b2920719ee381b440fbe, if you think it makes sense to do so. Or I can just create an issue to discuss some of the naming changes, `NOT NULL` for data column, etc. that can go into a future PR.

Bulat-Ziganshin:
thanks, now I see your point. I will continue code review

datastore/sqlite.nim
```diff
@@ -136,9 +142,26 @@ template journalModePragmaStmt*(env: SQLite): RawStmtPtr =
...
+  # see also https://www.sqlite.org/rescode.html
+
+  # in order to detect an out-of-memory error check that the result is a null
+  # pointer and that the result code is an error code
+  if x.isNil:
+    let
+      code = sqlite3_errcode(sqlite3_db_handle(s))
+
+    if not (code in [SQLITE_OK, SQLITE_ROW, SQLITE_DONE]):
+      raise (ref Defect)(msg: $sqlite3_errstr(code))
```
Bulat-Ziganshin: I don't think it's a Defect. Defect is an error in OUR code, it's fast-quit because it's supposed that we can find the bug, fix it and then ran the fixed version.

This error, though, like other errors returned by sqlite engine, is something internal to the engine and outside of our control. So, we should handle it similarly to situations when sqlite DB is broken, i.e. return failure.

datastore/sqlite_datastore.nim
```diff
@@ -33,53 +39,149 @@ type
...
   """
 
   createStmtStr = """
-    CREATE TABLE IF NOT EXISTS """ & tableTitle & """ (
-      id """ & IdType & """ NOT NULL PRIMARY KEY,
-      data """ & DataType & """,
-      timestamp """ & TimestampType & """ NOT NULL
+    CREATE TABLE IF NOT EXISTS """ & tableName & """ (
+      """ & idColName & """ """ & idColType & """ NOT NULL PRIMARY KEY,
+      """ & dataColName & """ """ & dataColType & """,
```
Bulat-Ziganshin, line 75: I think that dataCol should be NOT NULL too.

Accordingly, bindParam procedure should store empty blobs (val.len == 0) as empty strings (using sqlite3_bind_blob with ptr!=NULL and len=0). 

Add tests for store/retrieval of empty blocks if we don't yet have it.

Note that code still should handle NULL values stored in any column since we may deal with DB file created by other CREATE statement.

datastore/sqlite_datastore.nim
```diff
@@ -33,53 +39,149 @@ type
...
-    REPLACE INTO """ & tableTitle & """ (
-      id, data, timestamp
+    REPLACE INTO """ & tableName & """ (
+      """ & idColName & """,
+      """ & dataColName & """,
+      """ & timestampColName & """
     ) VALUES (?, ?, ?);
   """
 
+template checkColMetadata(s: RawStmtPtr, i: int, expectedName: string) =
```
Bulat-Ziganshin: It should be proc. Overall, use templates only when they can't be replaced by procs without much hassle (i.e. slightly more code with procs is OK)

datastore/sqlite_datastore.nim
```diff
@@ -33,53 +39,149 @@ type
...
+  let
+    colName = sqlite3_column_origin_name(s, i.cint)
+
+  if colName.isNil:
+    raise (ref Defect)(msg: "no column exists for index " & $i)
+
+  if $colName != expectedName:
+    raise (ref Defect)(msg: "original column name for index " & $i & " was \"" &
+      $colName & "\" but expected \"" & expectedName & "\"")
```
Bulat-Ziganshin: These msgs aren't very informative, I propose to include the SQL statement text

datastore/sqlite_datastore.nim
```diff
@@ -33,53 +39,149 @@ type
...
+        code = sqlite3_errcode(sqlite3_db_handle(s))
+
+      if not (code in [SQLITE_OK, SQLITE_ROW, SQLITE_DONE]):
+        raise (ref Defect)(msg: $sqlite3_errstr(code))
+
+    let
+      dataLen = sqlite3_column_bytes(s, i)
+
+    # an out-of-memory error can be inferred from a null pointer result
+    if (unsafeAddr dataLen).isNil:
```
Bulat-Ziganshin: sqlite3_column_bytes returns integer! Here you check the address of Nim variable in the stack the value is written to

datastore/sqlite_datastore.nim
```diff
@@ -33,53 +39,149 @@ type
...
+    $text.cstring
+
+proc dataCol*(
+  s: RawStmtPtr,
+  index = dataColIndex): BoundDataCol =
+
+  checkColMetadata(s, index, dataColName)
+
+  return proc (): seq[byte] =
+    let
```
Bulat-Ziganshin, line 129: I propose to convert the procedure code into proc sqlite3_get_blob_not_null and put it into sqlite.nim, so that the entire code here will be

```
  return proc (): seq[byte] =
    sqlite3_get_blob_not_null(s, index)
```
much like code in timestampCol.

Similarly, make function sqlite3_get_text_not_null for idCol

datastore/sqlite_datastore.nim
```diff
@@ -157,9 +259,12 @@ proc new*(
   # `pepare()` will fail and `new` will return an error with message
   # "SQL logic error"
 
+  let
+    getDataCol = dataCol(RawStmtPtr(getStmt), 0)
```
Bulat-Ziganshin: As I wrote in previous discussion, I propose to replace "0" here with getStmtDataCol constant defined at the next line after getStmtStr

datastore/sqlite_datastore.nim
```diff
@@ -181,37 +286,14 @@ proc close*(self: SQLiteDatastore) =
...
 method contains*(
   self: SQLiteDatastore,
   key: Key): Future[?!bool] {.async, locks: "unknown".} =
 
   var
     exists = false
 
-  proc onData(s: RawStmtPtr) {.closure.} =
+  proc onData(s: RawStmtPtr) =
     exists = sqlite3_column_int64(s, 0).bool
```
Bulat-Ziganshin: similarly, replace "0" with a constant defined next line after containsStmtStr

datastore/sqlite.nim
```diff
@@ -136,9 +142,26 @@ template journalModePragmaStmt*(env: SQLite): RawStmtPtr =
     s.dispose
     return failure $sqlite3_errstr(x)
 
-  if (let x = $sqlite3_column_text(s, 0).cstring; x != "memory" and x != "wal"):
+  let
+    x = sqlite3_column_text(s, 0).cstring
```
Bulat-Ziganshin: Use sqlite3_get_text_not_null here too (see below)

datastore/sqlite_datastore.nim
```diff
@@ -33,53 +39,149 @@ type
...
-  DataType = "BLOB"
-  TimestampType = "INTEGER"
-
   dbExt* = ".sqlite3"
-  tableTitle* = "Store"
+  tableName* = "Store"
+
+  idColName = "id"
+  dataColName = "data"
+  timestampColName = "timestamp"
```
Bulat-Ziganshin: I think that idColName, dataColName and timestampColName should be exported too, otherwise idCol... procedures aren't very useful

datastore/sqlite_datastore.nim
```diff
@@ -33,53 +39,149 @@ type
...
+
+  idColIndex = 0
+  dataColIndex = 1
+  timestampColIndex = 2
```
Bulat-Ziganshin: OTOH, these 3 consts probably should be removed completely. They make sense only for statements with `SELECT *` which are not used currently and can be replaced with `select id,name,timestamp` explicily listing all fields we want.

datastore/sqlite_datastore.nim
```diff
@@ -33,53 +39,149 @@ type
...
+  if colName.isNil:
+    raise (ref Defect)(msg: "no column exists for index " & $i)
+
+  if $colName != expectedName:
+    raise (ref Defect)(msg: "original column name for index " & $i & " was \"" &
+      $colName & "\" but expected \"" & expectedName & "\"")
+
+proc idCol*(
+  s: RawStmtPtr,
+  index = idColIndex): BoundIdCol =
```
Bulat-Ziganshin: I propose to drop default value in all 3 procs. It's useful only for `SELECT *` statements

datastore/sqlite.nim
michaelsbradleyjr: `Defect`

https://nim-lang.org/docs/system.html#Defect
> Abstract base class for all exceptions that Nim's runtime raises but that are strictly uncatchable as they can also be mapped to a `quit` / `trap` / `exit` operation. 

So in general we raise `Defect` when an error is unrecoverable, and by that I mean to imply the program state is in a bad way so that it doesn't make sense to e.g. log the error then ignore it and proceed.

*Aside:*  in Nim v1.2 you sometimes *can* catch a `Defect`, usually unintentionally, which is quite unfortunate and contrary to the docs. I think that bug is fixed in Nim v1.4 and later versions.

In this case it would work to do:
```nim
if not (code in [SQLITE_OK, SQLITE_ROW, SQLITE_DONE]):
  return failure $sqlite3_errstr(code)
```

The reason I chose to raise a `Defect` instead is that I considered the OOM error to be unrecoverable. If we `return failure` then the caller of `SQLiteDatastore.new` could make the choice to raise a `Defect` since the program likely can't proceed if object construction fails for whatever reason (OOM, bad path, etc.)

Any additional thoughts on this?

datastore/sqlite_datastore.nim
michaelsbradleyjr: We already have tests for storage and retrieval of empty data:

**put**
https://github.com/status-im/nim-datastore/blob/a82389743cc5e342b6ce64a9a79b025214d7572d/tests/datastore/test_sqlite_datastore.nim#L121
https://github.com/status-im/nim-datastore/blob/4e8a2d650621d3887844d9d31c9d4e9a6ee5577b/tests/datastore/test_sqlite_datastore.nim#L131-L136
(need to scroll vertically in the following code-box)

https://github.com/status-im/nim-datastore/blob/a82389743cc5e342b6ce64a9a79b025214d7572d/tests/datastore/test_sqlite_datastore.nim#L157-L171
**get**
https://github.com/status-im/nim-datastore/blob/a82389743cc5e342b6ce64a9a79b025214d7572d/tests/datastore/test_sqlite_datastore.nim#L302-L313

> I think that dataCol should be NOT NULL too.
>
> Accordingly, bindParam procedure should store empty blobs (val.len == 0) as empty strings (using sqlite3_bind_blob with ptr!=NULL and len=0).

I can make this change, but I'd like to understand the motivation better. What advantages are there over `NULL` representing empty data?

> Note that code still should handle NULL values stored in any column since we may deal with DB file created by other CREATE statement.

This worries me a bit, seems like it may overcomplicate things. We already have a strict requirement that a database schema created by a different `CREATE` statement must be equivalent to or a superset of our schema (leaving aside `NULL` consideration), else creation of the prepared statements in `SQLiteDatastore.new` will fail and therefore object construction will fail.

Could we simply document that a user should migrate data from their own database into one created by nim-datastore if they want to work with pre-existing data?

datastore/sqlite_datastore.nim
michaelsbradleyjr: Changed.

datastore/sqlite_datastore.nim
michaelsbradleyjr: What do you mean by *"include the SQL statement text"*?

To see this error message in action, keep in mind this custom query in the test:
```
SELECT timestamp as foo, id as baz, data as bar FROM ...
```
And change this line in the test:
https://github.com/status-im/nim-datastore/blob/06a94ec4cdebc794fed3ef1ad2603bc9edc21a25/tests/datastore/test_sqlite_datastore.nim#L147
So that it's `idCol = idCol(RawStmtPtr(prequery), 0)`.

When the test is run a Defect will be raised with error message:
> Unhandled exception: original column name for index 0 was "timestamp" but expected "id" [Defect]

If the author of a custom query compares that kind of error message with the `SELECT` statement they wrote, I hope it would be clear that they mismatched the index with the column order they specified in `SELECT ...`.


However, I realize I am biased because I am thinking about it in terms of what led me to have this metadata check and error message in the first place.

datastore/sqlite_datastore.nim
michaelsbradleyjr: Change made.

One or two of them could still be useful for `SELECT id FROM ...` and `SELECT id, data FROM ...`, or all of them for `SELECT id, data, timestamp FROM ...`, but the utility is admittedly low.

datastore/sqlite_datastore.nim
michaelsbradleyjr: Changed.

datastore/sqlite_datastore.nim
michaelsbradleyjr: Change made.

Given that the schema created by our `CREATE` statement is quite small, I imagined that most users writing custom queries would have it in mind and use names `id`, `data`, `timestamp` in their `SELECT` statements instead of combining strings with exported constants to build those statements.

However, I do think it's better to export them since careful users will make use of the constants to avoid problems in the case we change our schema.

datastore/sqlite_datastore.nim
michaelsbradleyjr: Change made.

datastore/sqlite_datastore.nim
michaelsbradleyjr: Change made.

datastore/sqlite_datastore.nim
Bulat-Ziganshin: I want to emphasize that indexes in `select *` and `select id` statements are different and even if they both equal the same 0 value - it's mere coincidence.

So if you have `select id`, you need to define index alongside this SQL text, rather than rely on hope to use the same field order as in `create` statement somewhere else and especially in a different module.

(of course, this makes more sense for larger packages and here looks like a mere chicanery, but still true)

datastore/sqlite_datastore.nim
michaelsbradleyjr: Fixed. 

After re-reading the docs just now I realized I misread re: `sqlite3_column_bytes()` in relation to `sqlite3_column_text/blob()`, and that the way I was checking it didn't make sense, as you noted.

datastore/sqlite_datastore.nim
Bulat-Ziganshin: Noone executes isoltaed SQL statements. Instead, they run their complex program and when it quits with message `expected id...`, they yet have to fugure out which program part did that. By adding this extra info, we can simplify the debugging process.

It's not mandatory, just an opportunity.

datastore/sqlite_datastore.nim
Bulat-Ziganshin: It's rather ironic that INSIDE the module you use idColName/... to build sql statements, but expect that external users will use fixed literals

datastore/sqlite_datastore.nim
michaelsbradleyjr: Change made. I realized what you meant after your last reply in this thread. It's a very good idea, thanks!

datastore/sqlite_datastore.nim
michaelsbradleyjr: I introduced constants for the column names because the same strings / string-fragments were appearing on multiple lines in multiples contexts in the same module, especially with the introduction of `idCol`, etc. That was the whole motivation.

Like I said, it *is* better to export those constants because careful users will want to use them, and next time I will remember that up front. I still anticipate that because the schema is small and/or because many external users will experiment/build custom queries with [DB Browser for SQLite](https://sqlitebrowser.org/) or the `sqlite3` command-line (that's what I would do), where there are no Nim constants but only names visible in the schema, those custom queries will end up embedding the column names when they're put into the external users' `.nim` modules. So I don't see anything ironic here, I think we just have different ways of thinking about these things, which combined will hopefully end up with a good result.

datastore/sqlite.nim
michaelsbradleyjr: I've tentatively introduced `sqlite3_column_text_not_null()` in `datastore/sqlite.nim` per b8e3b22f05c9a943370f4b19db66a0f65f32fd4f.

This naturally raises some concerns re: earlier discussions, mainly https://github.com/status-im/nim-datastore/pull/17#discussion_r920319710.

If `sqlite3_column_text_not_null()` were to not raise a Defect for an OOM error, and instead we propagated the failure via `return failure`, then measures to handle `Result[T, E]` will naturally spread outward, e.g. into `idCol` (hashtag  `#life-with-monads`).

If we're okay with `sqlite3_column_text_not_null()` raising a Defect for OOM error, then let's go with it. Otherwise, I suggest dropping commit b8e3b22f05c9a943370f4b19db66a0f65f32fd4f.

emizzle (approved):
LGTM. @Bulat-Ziganshin, what do you think about merging this as is and capturing any further requirements in a new issue? That way, we can move forward with the integration into Codex and manage scope better.

Bulat-Ziganshin:
@emizzle agree

datastore/sqlite.nim
Bulat-Ziganshin: OOM in Nim runtime itself is unrecoverable. OOM in external program is just a failure. But since SQLite engine is built into executable, I'm not sure.

Probably, it's better to be on the safe side and make it Defect. It probably will be very rare Defect.

emizzle:
> @emizzle agree

Thanks @Bulat-Ziganshin. If there are some requirements you feel that are not met in this PR, would you mind creating an issue for them please?

Issue #11, opened 2022-06-22
Refactor the Datastore API to be generic


michaelsbradleyjr:
This will allow more flexibility re: key and data parameters.

Because Nim's `method` facility doesn't support generics and because multi-methods are deprecated, we'll need to choose from one of several options:
* vtable approach as seen in e.g. nim-faststreams
* The approach taken in nim-eth's [kvstore.nim](https://github.com/status-im/nim-eth/blob/master/eth/db/kvstore.nim)
* Defining the core API in `datastore/datastore.nim` such that it is intended to be consumed by implementors of `Datastore` via `include` instead of `import`.

Note: the last option is currently considered the least desirable per previous team discussion, but I've included it in the list of possible approaches because it can work even if there are reasons to avoid it.

PR #19, opened 2022-07-20
add basic query implementation

### LINKED ISSUE #8: Add `query` implementation for listing blocks in a store

michaelsbradleyjr:
This PR adds a basic `query` implementation for base `Datastore`, `NullDatastore`, and most importantly `SQLiteDatastore`.

Closes #8.

An implementation for `FileSystemDatastore` will be skipped at this time, though it can be revisited in the future.

The immediate use case for this facility is https://github.com/status-im/nim-codex/issues/138, it's also relevant to https://github.com/status-im/nim-codex/issues/131.

For https://github.com/status-im/nim-codex/issues/137, if we want to use `TieredDatastore`, then some decisions need to made re: the desired behavior, as explained below.

**General Notes**

The first choice was *how* to implement the facility: should it be `method query` or `iterator query`? The second choice was whether it should be sync or async.

A `method query` approach would naturally involve `return`ing `seq[(key, data)]` (entire results for the query), while `iterator query` would involve `yield`ing `(key, data)` one at a time.

I chose to proceed with `iterator query` and made it async:

```nim
type
  QueryResponse* = tuple[key: Key, data: seq[byte]]

iterator query*(self: SQLiteDatastore, query: Query): Future[QueryResponse] =
  ...
```

Note that this is presently *"faux async"* just like `method get`, etc.

**Types `Query` and `QueryKey`**

The original vision of Python Datastore (embraced by various implementations in JavaScript and Go) is to have a `Query` class/object that provides a high-level almost declarative API for specifying what to query, how to filter, sort, and so on.

The basic implementation in this PR does **not** attempt to fulfill that vision. Rather, a `Query` object is constructed only with a `QueryKey`, and at the moment `QueryKey` is simply an alias for `Key`.

In the case of `SQLiteDatastore`, the idea is that a `QueryKey` will use the [`GLOB`](https://sqlite.org/lang_expr.html#glob) syntax supported by SQLite to match multiple rows.

Imagine that a Codex manifest with CID `abcd1234` is stored with key `manifest:abcd1234`. To retrieve all the blocks stored for that manifest, one could do something like:

```nim
for block in ds.query(Query.init("manifest:abcd1234/block:*")):
  let
    # key will look like "manifest:abcd1234/block:[cid]"
    (key, data) = await block
    
  ...
```

**Limitations of SQLite GLOB**

SQLite's GLOB syntax is *not* a form of regular expressions. The wildcard `*` is greedy such that the following won't work re: querying for all the manifests but not their blocks (per hypothetical example above):

```nim
ds.query(Query.init("manifest:*[^/]"))
```

That would match all manifests *and* all their blocks.

If desirable, there are a number of ways to work around that at the application level (i.e. in a program using `SQLiteDatastore`), or we could consider enabling PCRE support in SQLite and using `REGEXP` instead of `GLOB` for the `QueryKey` parameter.

**Async + Iterator**

Consider the support built into the JavaScript language for asynchronous iteration:
* [`Symbol.asyncIterator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncIterator)
* [`for-await...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of)
* [`async function*`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function*)

Those language constructs work together so that async iteration is well-defined and distinct from sync iteration. For example, `for-await...of` needs to be used in the general case of async iteration because it's not generally possible to know in advance how many times an async iterable will yield a value. In other words, you can't use `Promise.all()`, you need to step through each promised value, awaiting each in turn. Or you can use some sugar lib like [IxJS](https://github.com/ReactiveX/IxJS#readme) to avoid manually writing `for-await...of` loops.

We don't currently have libs or lang features for Nim that are dedicated to async iteration. So, it's important to understand that it will be up to the caller to `await` on each step in a loop over async `iterator query`, at least in the general case, i.e. instead of trying to collect all the responses into chronos' `allFinished()`.

Because `iterator query` for `SQLiteDatastore` is *"faux async"*, it technically could work to exhaust the iterator and run the collected `Future`s through `allFinished()`. But when we move to a case like `TieredDatastore` (depending on how that works), or when we have a multi-threaded solution where the actual querying of the database takes place on a worker thread, then it's not feasible to do that, at least not without additional machinery.

**TieredDatastore**

In the original Python Datastore, there is this [doc comment](https://github.com/datastore/datastore/blob/7ccf0cd4748001d3dbf5e6dda369b0f63e0269d3/datastore/core/basic.py#L1027-L1028) for `class TieredDatastore`:
```
Datastores should be arranged in order of completeness,
with the most complete datastore last, as it will handle query calls.
```
And below that:
```python
def query(self, query):
    '''Returns a sequence of objects matching criteria expressed in `query`.
    The last datastore will handle all query calls, as it has a (if not
    the only) complete record of all objects.
    '''
    # queries hit the last (most complete) datastore
    return self._stores[-1].query(query)
```
Indeed, it's not clear how a `query` implementation (in Nim, Python, or other lang) should operate internally if a query op were to span multiple stores.

So `iterator query` for nim-datastore's `TieredDatastore` should probably have the last datastore directly handle queries.

However, that does not mean that application-specific derivatives of `TieredDatastore` could not have specialized logic for e.g. `method get` vs. `iterator query`.

Codex, for example, might have this arrangement re: nim-datastore layers in a derivative of `TieredDatastore`:
```
LRU cache -> SQLite -> Network
```
While `method get` may attempt to resolve a key from `Network` if neither `LRU cache` nor `SQLite` has it, `iterator query` could bypass `LRU cache` and go directly to `SQlite` while also *not* attempting to fall through to `Network`.

**TODO in future PRs**

We could prevent GLOB characters (`*`, `?`, `[`, `]`) from being used in strings passed to `Key.init()` while allowing them to be used in `QueryKey.init()`.

In #11 we look toward a refactor that will be generic with respect to parameters and return types. That will provide additional flexibility, allowing concrete implementations to diverge in terms of their choice of base `Key` and other base types, how e.g. `Key` relates to `Query` and `QueryKey` (and even data layout in the backend), and making it simpler for applications to specialize on types and behaviors for `get`, `query`, etc. for a specific backend on which they choose to build.

### DIFF
--- a/datastore/sqlite_datastore.nim
+++ b/datastore/sqlite_datastore.nim
@@ -6,7 +6,7 @@ import pkg/questionable
 import pkg/questionable/results
 import pkg/sqlite3_abi
 import pkg/stew/byteutils
-from pkg/stew/results as stewResults import get, isErr
+from pkg/stew/results as stewResults import isErr
 import pkg/upraises
 
 import ./datastore
@@ -33,6 +33,8 @@ type
   GetStmt = SQLiteStmt[(string), void]
 
   PutStmt = SQLiteStmt[(string, seq[byte], int64), void]
+
+  QueryStmt = SQLiteStmt[(string), void]
 
   SQLiteDatastore* = ref object of Datastore
     dbPath: string
@@ -97,6 +99,14 @@ const
     ) VALUES (?, ?, ?);
   """
 
+  queryStmtStr = """
+    SELECT """ & idColName & """, """ & dataColName & """ FROM """ & tableName &
+    """ WHERE """ & idColName & """ GLOB ?;
+  """
+
+  queryStmtIdCol = 0
+  queryStmtDataCol = 1
+
 proc checkColMetadata(s: RawStmtPtr, i: int, expectedName: string) =
   let
     colName = sqlite3_column_origin_name(s, i.cint)
@@ -140,10 +150,10 @@ proc dataCol*(
     # result code is an error code
     if blob.isNil:
       let
-        code = sqlite3_errcode(sqlite3_db_handle(s))
+        v = sqlite3_errcode(sqlite3_db_handle(s))
 
-      if not (code in [SQLITE_OK, SQLITE_ROW, SQLITE_DONE]):
-        raise (ref Defect)(msg: $sqlite3_errstr(code))
+      if not (v in [SQLITE_OK, SQLITE_ROW, SQLITE_DONE]):
+        raise (ref Defect)(msg: $sqlite3_errstr(v))
 
     let
       dataLen = sqlite3_column_bytes(s, i)
@@ -339,8 +349,64 @@ method put*(
 
   return await self.put(key, data, timestamp())
 
-# method query*(
-#   self: SQLiteDatastore,
-#   query: ...): Future[?!(?...)] {.async, locks: "unknown".} =
-#
-#   return success ....some
+iterator query*(
+  self: SQLiteDatastore,
+  query: Query): Future[QueryResponse] =
+
+  let
+    queryStmt = QueryStmt.prepare(
+      self.env, queryStmtStr).expect("should not fail")
+
+    s = RawStmtPtr(queryStmt)
+
+  defer:
+    discard sqlite3_reset(s)
+    discard sqlite3_clear_bindings(s)
+    s.dispose
+
+  let
+    v = sqlite3_bind_text(s, 1.cint, query.key.id.cstring, -1.cint,
+      SQLITE_TRANSIENT_GCSAFE)
+
+  if not (v == SQLITE_OK):
+    raise (ref Defect)(msg: $sqlite3_errstr(v))
+
+  while true:
+    let
+      v = sqlite3_step(s)
+
+    case v
+    of SQLITE_ROW:
+      let
+        key = Key.init($sqlite3_column_text_not_null(
+          s, queryStmtIdCol)).expect("should not fail")
+
+        blob = sqlite3_column_blob(s, queryStmtDataCol)
+
+      # detect out-of-memory error
+      # see the conversion table and final paragraph of:
+      # https://www.sqlite.org/c3ref/column_blob.html
+      # see also https://www.sqlite.org/rescode.html
+
+      # the "data" column can be NULL so in order to detect an out-of-memory
+      # error it is necessary to check that the result is a null pointer and
+      # that the result code is an error code
+      if blob.isNil:
+        let
+          v = sqlite3_errcode(sqlite3_db_handle(s))
+
+        if not (v in [SQLITE_OK, SQLITE_ROW, SQLITE_DONE]):
+          raise (ref Defect)(msg: $sqlite3_errstr(v))
+
+      let
+        dataLen = sqlite3_column_bytes(s, queryStmtDataCol)
+        dataBytes = cast[ptr UncheckedArray[byte]](blob)
+        data = @(toOpenArray(dataBytes, 0, dataLen - 1))
+        fut = newFuture[QueryResponse]()
+
+      fut.complete((key, data))
+      yield fut
+    of SQLITE_DONE:
+      break
+    else:
+      raise (ref Defect)(msg: $sqlite3_errstr(v))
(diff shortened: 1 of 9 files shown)


datastore/sqlite_datastore.nim
```diff
@@ -339,8 +349,64 @@ method put*(
...
+
+      # the "data" column can be NULL so in order to detect an out-of-memory
+      # error it is necessary to check that the result is a null pointer and
+      # that the result code is an error code
+      if blob.isNil:
+        let
+          v = sqlite3_errcode(sqlite3_db_handle(s))
+
+        if not (v in [SQLITE_OK, SQLITE_ROW, SQLITE_DONE]):
+          raise (ref Defect)(msg: $sqlite3_errstr(v))
```
Bulat-Ziganshin, line 399: This became repetitive pattern, may be replaced sometime in future

datastore/sqlite_datastore.nim
```diff
@@ -339,8 +349,64 @@ method put*(
...
+    discard sqlite3_reset(s)
+    discard sqlite3_clear_bindings(s)
+    s.dispose
+
+  let
+    v = sqlite3_bind_text(s, 1.cint, query.key.id.cstring, -1.cint,
+      SQLITE_TRANSIENT_GCSAFE)
+
+  if not (v == SQLITE_OK):
+    raise (ref Defect)(msg: $sqlite3_errstr(v))
```
Bulat-Ziganshin, line 372: As I said before, we should discuss when raise Defect and when return failure, but at this moment let's accept the already implemented approach

datastore/sqlite_datastore.nim
```diff
@@ -339,8 +349,64 @@ method put*(
...
+          v = sqlite3_errcode(sqlite3_db_handle(s))
+
+        if not (v in [SQLITE_OK, SQLITE_ROW, SQLITE_DONE]):
+          raise (ref Defect)(msg: $sqlite3_errstr(v))
+
+      let
+        dataLen = sqlite3_column_bytes(s, queryStmtDataCol)
+        dataBytes = cast[ptr UncheckedArray[byte]](blob)
+        data = @(toOpenArray(dataBytes, 0, dataLen - 1))
+        fut = newFuture[QueryResponse]()
```
Bulat-Ziganshin, line 405: will it work with blob==nil ? your testing code for queries doesn't include empty blobs

Bulat-Ziganshin (commented):
I accept this PR, except for the need to test queries with empty blobs returned. Otherwise, it's OK for merge

datastore/sqlite_datastore.nim
michaelsbradleyjr: In this case I wasn't sure of the best approach because we can't `return` from an inline iterator. We could `yield` a failure and `break` the internal loop, or leave it to the user to `break` in their loop over the iterator when an iteration-step yields a failure. Either way the cleanup in the `defer` block will be triggered.

Since I wasn't sure, for now I chose to raise Defect on any error, either directly or via nim-result's `expect`.

datastore/sqlite_datastore.nim
michaelsbradleyjr: I'll add some tests for that. Should already work since the approach is roughly the same for `method get`, which supports empty blobs.

michaelsbradleyjr:
> test queries with empty blobs returned. Otherwise, it's OK for merge

I've accounted for empty blobs in 32596eb.

I also replaced the original PR description with notes that I thought to be relevant now and in the future, even if at present we move forward with the PR as-is.

Let me know what you think, and if we're good to go, I'll squash my `[wip]` commits (really this time 🤦) and then merge it.

datastore/sqlite_datastore.nim
michaelsbradleyjr: I've accounted for empty blobs in 32596eb.

Bulat-Ziganshin:
your PR comments should be documents of its own.

As far as you tested empty blobs, this PR is OK for me.

as of chosen API, it should be debated before implementation. Now the API you choosen is already implemented, so let's Merge it and go to work on integration

PR #23, opened 2022-08-04
refactor Datastore impls so root/basePath creation is user's responsibility

### LINKED ISSUE #21: Datastore impls should not create directories
Their constructors should only check whether the specified directory exists and return a failure if it does not.

See https://github.com/status-im/nim-codex/pull/160#discussion_r933496748.

michaelsbradleyjr:
Closes #21.

### DIFF
--- a/datastore/filesystem_datastore.nim
+++ b/datastore/filesystem_datastore.nim
@@ -21,18 +21,17 @@ const
 
 proc new*(
   T: type FileSystemDatastore,
-  root = "data"): ?!T =
+  root: string): ?!T =
 
   try:
     let
       root = if root.isAbsolute: root
              else: getCurrentDir() / root
 
-    createDir(root)
-    success T(root: root)
-
-  except IOError as e:
-    failure e
+    if not dirExists(root):
+      failure "directory does not exist: " & root
+    else:
+      success T(root: root)
 
   except OSError as e:
     failure e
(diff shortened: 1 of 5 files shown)


datastore/filesystem_datastore.nim
```diff
@@ -21,18 +21,17 @@ const
 
 proc new*(
   T: type FileSystemDatastore,
-  root = "data"): ?!T =
+  root = getCurrentDir()): ?!T =
```
michaelsbradleyjr: @dryajov `getCurrentDir()` seems like a reasonable default value, but would it be preferable to not have a default value?

datastore/sqlite_datastore.nim
```diff
@@ -172,7 +172,7 @@ proc timestampCol*(
 
 proc new*(
   T: type SQLiteDatastore,
-  basePath = "data",
+  basePath = getCurrentDir(),
```
michaelsbradleyjr: Same question as in my previous comment.

emizzle:
Per meeting: should be a required param to supply a default directory

datastore/filesystem_datastore.nim
michaelsbradleyjr: Based on discussion during today's team call, I have changed it so there's no default value for the `root` parameter.

datastore/sqlite_datastore.nim
michaelsbradleyjr: Based on discussion during today's team call, I have changed it so there's no default value for the `basePath` parameter.

dryajov (approved):
LGTM

PR #25, opened 2022-09-04
make page size configurable in SQLiteDatastore.new

### LINKED ISSUE #6: Make page size configurable in `SQLiteDatastore.new`
Default should be `4096`, same as SQLite's default.

michaelsbradleyjr:
Also make cache size and journal mode configurable.

Closes #6.
Related to status-im/nim-codex#234.

### DIFF
--- a/datastore/sqlite.nim
+++ b/datastore/sqlite.nim
@@ -1,3 +1,6 @@
+import std/math
+import std/unicode
+
 import pkg/questionable
 import pkg/questionable/results
 import pkg/sqlite3_abi
@@ -19,6 +22,10 @@ type
     val*: T
 
   DataProc* = proc(s: RawStmtPtr) {.closure, gcsafe.}
+
+  # https://www.sqlite.org/pragma.html#pragma_journal_mode
+  JournalMode* = enum
+    DELETE, TRUNCATE, PERSIST, MEMORY, WAL, OFF
 
   NoParams* = tuple # empty tuple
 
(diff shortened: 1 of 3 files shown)


Bulat-Ziganshin:
Looks great for me

datastore/sqlite.nim
```diff
@@ -172,9 +179,67 @@ proc sqlite3_column_text_not_null*(
...
+    s1 = prepare(env, "PRAGMA page_size = " & $pageSize & ";")
+
+  s1.dispose
+
+  let
+    s2 = prepare(env, "PRAGMA page_size;")
+
+  if (let x = sqlite3_step(s2); x != SQLITE_ROW):
+    s2.dispose
+    return failure $sqlite3_errstr(x)
```
emizzle, line 197: Probably not relevant to this PR, but how does a `Result[T, string]` equate to the return type of `RawStmtPtr`?

emizzle (approved):
LGTM, nice work! I just had one question but likely isn't related to this PR.

datastore/sqlite.nim
michaelsbradleyjr: It's a template, so gets expanded into the body of `proc new*(T: type SQLiteDatastore, ...): ?!T`, and in the case of `pageSizePragmaStmt(...)` the returned `RawStmtPtr` is assigned to `let pageSizePragmaStmt = ...`, for which that type is the desired one.

Bulat-Ziganshin:
From Michael's message:

There are some other pragmas (possibly some compile-time options) we may want to consider re: improving nim-codex's performance. I'm no expert, just becoming more aware of what's possible after spending more time with SQLite's docs and following links therein.

https://www.sqlite.org/pragma.html#pragma_wal_autocheckpoint
https://www.sqlite.org/wal.html#ckpt
https://www.sqlite.org/compile.html#default_wal_autocheckpoint

https://www.sqlite.org/pragma.html#pragma_mmap_size
https://www.sqlite.org/mmap.html

It's pretty straightforward to add support for other pragmas; doesn't necessarily have to be in a PR, could just be on a branch in status-im/nim-datastore for sake of experimentation.

Bulat-Ziganshin:
I think that it should have different logic - if sqlite store already exists and has different pagesize (and other parameters), it should continue to work. Ideally, it should print a notice about that, but don't stop the work.

For a starter, we may just comment out code that checks parameters after setting them.

Current behavior means that unlucky user will have to guess DB parameters before he can use this DB :)

PR #26, opened 2022-09-10
simplify key api

dryajov:


c-blake:
FWIW, I like these changes.

PR #27, opened 2022-09-12
introduce methods close and query in the base API and refactor accordingly

### LINKED ISSUE #28: Implement methods close and query so TieredDatastore can work properly

michaelsbradleyjr:
The goal of this PR is to make `Datastore` and derived types more flexible, e.g. in nim-libp2p-dht we want to specify `providers: Datastore` instead of `providers: SQLiteDatastore` and have the freedom to choose one implementation or another without having to change the type definition (see https://github.com/status-im/nim-libp2p-dht/pull/41#discussion_r966316669).

`method close` is introduced in a similar manner as was done for `BlockStore` in https://github.com/status-im/nim-codex/pull/160.

`method query` required a bit more experimentation, but the end result seems viable. The challenge is that Nim's `iterator` does not support the dynamic dispatch of `method`. However, we can specify a single return type (see `type QueryImpl` in this PR) for implementations of `method query`, and then in the returned `queryImpl` iterator use object conversion to get back to the correct specific type derived from `Datastore`.

I did a small amount of additional cleanup, e.g. it's not necessary to specify the `async` pragma in base methods of `Datastore`.

In the tests for `SQLiteDatastore`, in some places I had to change the name of a variable `key` to something else (I chose `rkey`) otherwise I was getting compile-time errors from clang/gcc:
```
CC: datastore/test_sqlite_datastore.nim
/path/to/.cache/nim/test_all_d/@mdatastore@stest_sqlite_datastore.nim.c:1087:39: error: duplicate member 'key89'
tyObject_Key__6b3P9aUji2WzyTKPwAJWNUQ key89;
                                      ^
/path/to/.cache/nim/test_all_d/@mdatastore@stest_sqlite_datastore.nim.c:1007:39: note: previous declaration is here
tyObject_Key__6b3P9aUji2WzyTKPwAJWNUQ key89;
                                      ^
1 error generated.
Error: execution of an external compiler program 'clang -c  -w -ferror-limit=3 -pthread -DSQLITE_ENABLE_COLUMN_METADATA   -I'/path/to/.choosenim/toolchains/nim-#devel/lib' -I/path/to/repos/nim-datastore/tests -o /path/to/.cache/nim/test_all_d/@mdatastore@stest_sqlite_datastore.nim.c.o /path/to/.cache/nim/test_all_d/@mdatastore@stest_sqlite_datastore.nim.c' failed with exit code: 1
```
I've seen similar previously, I believe it's owing to a bug/limitation in Nim's compiler.

---

Closes #28.

### DIFF
--- a/tests/datastore/test_tiered_datastore.nim
+++ b/tests/datastore/test_tiered_datastore.nim
@@ -30,7 +30,7 @@ suite "TieredDatastore":
     ds2 = FileSystemDatastore.new(rootAbs).get
 
   teardown:
-    if not ds1.isNil: ds1.close
+    if not ds1.isNil: await ds1.close
     ds1 = nil
     removeDir(rootAbs)
     require(not dirExists(rootAbs))
@@ -132,7 +132,7 @@ suite "TieredDatastore":
       (await ds1.get(key)).get.get == bytes
       (await ds2.get(key)).get.get == bytes
 
-    ds1.close
+    await ds1.close
     ds1 = SQLiteDatastore.new(memory).get
     ds = TieredDatastore.new(ds1, ds2).get
 
--- a/tests/datastore/test_tiered_datastore.nim
+++ b/tests/datastore/test_tiered_datastore.nim
@@ -1,3 +1,4 @@
+import std/algorithm
 import std/options
 import std/os
 
@@ -19,21 +20,21 @@ suite "TieredDatastore":
     rootAbs = getCurrentDir() / root
 
   var
-    ds1: SQLiteDatastore
-    ds2: FileSystemDatastore
+    ds1: FileSystemDatastore
+    ds2: SQLiteDatastore
 
   setup:
     removeDir(rootAbs)
     require(not dirExists(rootAbs))
     createDir(rootAbs)
-    ds1 = SQLiteDatastore.new(memory).get
-    ds2 = FileSystemDatastore.new(rootAbs).get
+    ds1 = FileSystemDatastore.new(rootAbs).get
+    ds2 = SQLiteDatastore.new(memory).get
 
   teardown:
-    if not ds1.isNil: await ds1.close
-    ds1 = nil
     removeDir(rootAbs)
     require(not dirExists(rootAbs))
+    if not ds2.isNil: await ds2.close
+    ds2 = nil
 
   asyncTest "new":
     check:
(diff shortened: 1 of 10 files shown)


tests/datastore/test_tiered_datastore.nim
```diff
@@ -149,6 +152,55 @@ suite "TieredDatastore":
...
+
+      key1 = Key.init("a/b").get
+      key2 = Key.init("a/b:c").get
+      key3 = Key.init("a/b:c/d").get
+
+      bytes1  = @[1.byte, 2.byte, 3.byte]
+      bytes2  = @[4.byte, 5.byte, 6.byte]
+      bytes3: seq[byte] = @[]
+
+      queryKey1 = Key.init("a/*").get
```
dryajov, line 167: I don't think there is such a thing as a `*` glob, in the key semantics. A key already has hierarchy implied and allows `/a` would mean to match all the keys with prefix `a`, a.k.a do a `*` match.

michaelsbradleyjr: That goes back to #19. The key semantics are a bit underspecified in various respects, and adopting SQLite's GLOB syntax seemed natural. I was going to add some review comments on #26 but it's already been merged, so I'll create an issue, and/or we can discuss in discord, etc.

dryajov: The key semantics are pretty well defined, which is what the query interface uses, there are no special semantics outside of the key itself + some filtering facilities. The original python implementation and it's derivations take a very straightforward approach to deriving this semantics, which are - 

(pseudocode)

- Keys are hierarchical, and each level of the key gets more and more specific 
  - `/a` - `select * from <table> where key like '/a'%`
  - `/a/b` - `select * from <table> where key like '/a/b'%`
- You can specify keys with namespaces `/a/[field]:[name]`
  - `/a/name:Bob` - `select * from <table> where key like '/a/name:Bob'%`

They query interface then allows you to add filters. 

Consider a key such as - `/a/b/name:Bob/age:33`, to filter by name and age we would do something like:

- Key.init('/a/b') and Filter.init('name', 'Bob', '==') and Filter.init('age', '30', '=>')
  - This roughly translates into 
    - `select * from <table> where name = 'Bob' and age => 30 (select substr(...) as name, substr(...) as age from <table> where key like '/a/b'%)`

PR #31, opened 2022-09-20
Better fsds path sanitization

dryajov:


dryajov:
I need to setup a windows env to fix the windows related issues.

datastore/fsds.nim
```diff
@@ -126,6 +136,12 @@ method put*(
   let
     path = self.path(key)
 
+  if not self.isRootSubdir(path):
+    return failure "Path is outside of `root` directory!"
+
+  if not path.extractFilename.isValidFilename:
```
Bulat-Ziganshin: This line may raise an exception when key is "/" or "\\" since in this case 
1. the constructed path ends up with directory separator, f.e. `C:\Codex\repos\nim-datastore\tests\datastore\tests_data\`
2. extractFilename on such path returns empty string
3. isValidFilename fails on empty string

isValidFilename bug was fixed on Oct 21 2021 in line 3518 of this patch 
https://github.com/nim-lang/Nim/commit/62701cd3b92677c2014cf70b1cf5a5ba6e0468bf#diff-8ed10106605d9e0e3f28a927432acd8312e96791c96dbb126a52a7010cf4b44aL3517-L3518

This change was committed to devel branch on 20 Oct 2021, but seems not included even in Nim 1.6.6. BTW it also marks isValidFilename as deprecated.

The fix is to check that path.extractFilename isn't empty (in `contains` we need to check that prir to `addFileExt(FileExt)`

PR #45, opened 2023-09-07
Fix conflicting unit test

dryajov:


elcritch:
Thanks! I couldn't figure this out. BTW, should we just disable the Nim 1.2 builds?

elcritch (approved):

elcritch:
@dryajov fix this by removing the nim 1.2 runs. It seems datastore is mostly codex specific so there's no req's to support 1.2?

dryajov:
> @dryajov fix this by removing the nim 1.2 runs. It seems datastore is mostly codex specific so there's no req's to support 1.2?

Yeah, we don't need 1.2 support.

PR #48, opened 2023-09-15
Threaded IO datastore with minimal modifications

dryajov:


datastore/threads/threadproxyds.nim
```diff
@@ -0,0 +1,338 @@
...
+  var
+    res = ThreadResult[DataBuffer]()
+    ctx = TaskCtx[DataBuffer](
+      ds: addr self.ds,
+      res: addr res,
+      semaphore: addr self.semaphore,
+      signal: signal)
+
+  proc runTask() =
+    self.tp.spawn getTask(addr ctx, unsafeAddr key)
```
elcritch: Passing address of the `key` can lead to memory corruption issues. If the Future from the `get` method is cancelled it could free the memory before the `getTask` finishes. This is the same for `put` and `query` as well.

elcritch: I think the same issue would apply to `TaskCtx` as well. The object will be stored in the future.

datastore/threads/semaphore.nim
```diff
@@ -0,0 +1,58 @@
+import std/atomics
+import std/locks
+
+type
+  Semaphore* = object
+    count: int
+    size: int
+    lock {.align: 64.}: Lock
+    cond: Cond
```
elcritch, line 9: I like the semaphore idea. Though wouldn't Chronos have an async semaphore mechanism?

tests/datastore/testthreadproxyds.nim
```diff
@@ -0,0 +1,80 @@
...
+    taskPool = Taskpool.new(countProcessors() * 2)
+    ds = ThreadDatastore.new(sqlStore, taskPool).tryGet()
+
+  teardown:
+    (await ds.close()).tryGet()
+    taskPool.shutdown()
+
+  queryTests(ds, true)
+
+# TODO: needs a per key lock to work
```
elcritch: Yah, I think it'd be possible to do something like my `signalpool` for the key-lock. It's pretty much the same pattern just without pre-allocating the keys and doing the await on whether the key is used.

datastore/threads/threadproxyds.nim
```diff
@@ -0,0 +1,338 @@
...
+  ctx: ptr TaskCtx,
+  key: ptr Key) =
+  ## Run get in a thread task
+  ##
+
+  defer:
+    discard ctx[].signal.fireSync()
+    ctx[].semaphore[].release()
+
+  ctx[].semaphore[].acquire()
```
elcritch: Nice idea, this would keep the backend synchronous. 🤔 

Though it wouldn't provide back-pressure to the async side as async-requests could just keep queuing up no? Still need something to tie to the main chronos thread. 

On line 224, `signal = ThreadSignalPtr.new().valueOr` an errant async proc could still try to load say a 10,000 keys creating a new AsyncFD for each.

datastore/threads/threadproxyds.nim
dryajov: The object is stored in the closure, which is alive passed the lifetime of the future, I do cancel the features on shutdown and it seems to work fine for now.

datastore/threads/semaphore.nim
dryajov: There is one in libp2p and I actually used it before switching to the threaded version, I think tho, for the proxy the async semaphore should be fine.

datastore/threads/threadproxyds.nim
dryajov: Yeah, good point, I think in general we should probably revert back to the async semaphore, if you look at the history, I was using that previously.

dryajov: Tho, the `acquire` should happen before the thread signal.

tests/datastore/testthreadproxyds.nim
dryajov: It's actually much simpler, we don't need to do this synchronization at the thread level, same as with the semaphore, I realized that after I wrote the threaded semaphore. We just need a table where we store a future (any future), for the that the async `get/set/has/put/del` check and set, before spawning the task. Since we have an SPMC flow, we can do away with thread level locking in most places, this will also serialize read/write operations since the "locking" will happen at the future level and not at the thread level. 

I'll integrate this shortly, it sounds like a relatively straightforward thing to do.

datastore/threads/threadproxyds.nim
dryajov: Added cancellation support in the threaded tasks

elcritch: > The object is stored in the closure, which is alive passed the lifetime of the future, I do cancel the features on shutdown and it seems to work fine for now.

It is stored in the closure, but you can't count on the lifetime of that closure to exceed that of the threaded task. 

It may sorta work, as likely the GC isn't being run after you cancel the Future. But it's a fundamentally flawed approach, and will likely lead to bad behavior or worse corrupted memory when it's running for a while given our use cases. With threading it's not a matter of if, but when it'll happen.

It'd be possible to do a cancel check in the threaded tasks using an atomic flag but it'd be hard to get right with threading. At that point it's easier just to use the shared ptr.

There's another way by putting requests onto a queue and the results into another queue as well. But that's trickier to tie back to individualmfutures.

dryajov: > flawed approach,

As long as the closure context is alive, the memory for that closure is also alive, this is guaranteed by the chronos signaling primitive and the future that is attached to it. As long as the future is around it is safe to keep a _read only_ reference to it. This is the reason I reworked your implementation initially, because the amount of copying seemed excessive.

dryajov: but, if this becomes problematic, we can always copy instead...

dryajov:
Added some tests and a very ugly locking mechanism that we can hopefully refine. The good thing is that all tests pass and there are 0 changes to the api.

datastore/threads/threadproxyds.nim
elcritch: > this is guaranteed by the chronos signaling primitive and the future that is attached to it

That's a good point, and if so you'd be correct. 

Though it seems likely that if the datastore future were cancelled, it'd also cause that Future to be cancelled / freed as well. 

I'll look into it, because it would be nice, but it'd depend if Chronos kept the Future tied to the AsyncFD. If it cleans up those then they'll be freed then.

> but, if this becomes problematic, we can always copy instead... 

The problem there is that it would be *very* hard to track down. It's the sorta thing that'd plague tracking down and spending months trying to figure out.

Not trying to be a pain, but I've run into stuff like this with multi-threading before and it just makes things unreliable pretty quickly. And it almost always seems to happen at the worst time. 

We could skip the copy and take the pointer more safely if we did a lock that's released when the future is cancelled. The thread-task would need to acquire the lock when it was using the data (read or write). Then the cancel proc would need to block until the lock was free again. That would block the Chronos thread during that time, but it should only occur when a cancellation happens which should be rare.

Another way would be to change the FSDS / SQlite to use `DataBuffer` directly. Then you only have one copy. I also saw lots of GC errors yesterday when I ran tests with more threads. It was all in the Key conversion parts and iterators, after the data had been copied out. It was trying to allocate new strings and segfaulting. So we may need to setup the DS backends to use SharedPtr's or similar intenally to fix that if we can't figure out why the GC is breaking there.

dryajov: > Though it seems likely that if the datastore future were cancelled, it'd also cause that Future to be cancelled / freed as well.

It depends, we only cancel on datastore close and as such it should not happen unless explicitly triggered on app shutdown.

There is of course a few things missing and I'm not sure I cover all the code paths atm, but there shouldn't be any issues we couldn't possibly address with this setup - you can already see that I handle cancellations with an async monitor.

> Not trying to be a pain, but I've run into stuff like this with multi-threading before and it just makes things unreliable pretty quickly. And it almost always seems to happen at the worst time.

Overall our flow is pretty simple, there is no intertask data sharing at the thread level and we're constantly in a SPMC mode, which makes things a bit more straightforward.

I don't think there are any complex flows (and there hardly are for IO threads) in our setups, which again should make things a bit more straightforward. 

For the most part, if you take a look at nimbus it uses a similar approach with passing pointers to closure objects around. We might have one or two weird cases (I fixed one in the fs store query iterator), but overall there doesn't seem to be a need to overcomplicate things at this stage.

I do think that taskpool is probably not the best option for our setup - it was designed with CPU bound tasks in mind and it doesn't have any thread-scaling which might be beneficial in our case - but it should do for now.

Ideally we will eventually settle on something like this https://zig.news/kprotty/resource-efficient-thread-pools-with-zig-3291 or at the  very least https://github.com/libuv/libuv/blob/v1.x/src/threadpool.c

elcritch: Yet another way, we could use your original idea of a pool. If the main Chronos thread did book-keeping on it and owned all of the seq's we could rely on it to keep the lifetimes.

There'd need to be a flag for if thread-task was still using it. When the thread-task is done they mark it as released.

The main Chronos side would then need do some sort of bookkeeping on the Future's side. The `getTask` could possibly look for any "released" seqs of big enough size and use one of them. It'd probably also need to be able to request new ones as well. 

Hmmm, that might also provide the key-locking mechanism too.

elcritch: > There is of course a few things missing and I'm not sure I cover all the code paths atm, but there shouldn't be any issues we couldn't possibly address with this setup - you can already see that I handle cancellations with an async monitor.

The more likely source of cancellations would be defects or errors in other places which could lead to a chain that cancels the request. 

> Overall our flow is pretty simple, there is no intertask data sharing at the thread level and we're constantly in a SPMC mode, which makes things a bit more straightforward.

That's true, and Nimbus has an even simpler flow. 

> For the most part, if you take a look at nimbus it uses a similar approach with passing pointers to closure objects around. We might have one or two weird cases (I fixed one in the fs store query iterator), but overall there doesn't seem to be a need to overcomplicate things at this stage.

No I still don't agree given the cost vs risks / downsides. Adding a mechanism to handle it properly isn't that complex and there's a "minimum complexity" to handle these things well. There's already enough corner cases in Codex itself too -- this would add one more potential piece to debug.

It's something I would not feel comfortable signing off on a PR for. It's a couple of days and a few dozen lines of code at most to do a no-copy version, or already done if we leave the extra copies. 

It avoids too many potential pitfalls to justify it in my experience even with a fairly simple flow. Perhaps it'd work fine, but it's a core mechanism of Codex and skipping it doesn't really buy anything except future work and possible issues. It also slows down development because any testing failures I find would need to be verified it wasn't due to this. It'd be better to do it properly, then move on to other items that need worked on like a custom pool.

> I do think that taskpool is probably not the best option for our setup - it was designed with CPU bound tasks in mind and it doesn't have any thread-scaling which might be beneficial in our case - but it should do for now.

Yes that's true. It's not difficult to make a basic pool, and even a more advanced pool isn't too hard.

dryajov: > It's something I would not feel comfortable signing off on a PR for. It's a couple of days and a few dozen lines of code at most to do a no-copy version, or already done if we leave the extra copies.

If we decide to copy it's just a few lines of code worth of changes, I don't think this is a major issue and we can try both ways. For cancellations, I think we have a good way of handling them in the code right now, we just have to make sure we've covered all the code paths.

elcritch: > If we decide to copy it's just a few lines of code worth of changes, I don't think this is a major issue and we can try both ways. For cancellations, I think we have a good way of handling them in the code right now, we just have to make sure we've covered all the code paths.

Okay if we cover all the cancellation pathways that'd be fine. For me it's an important building block and getting the rest of the pieces can be hard enough by themselves. It's why I focus on them. I'll read a bit more closely what you did for the cancellation pieces.

elcritch:
> Added some tests and a very ugly locking mechanism that we can hopefully refine. The good thing is that all tests pass and there are 0 changes to the api.

Ok, nice.

datastore/threads/threadproxyds.nim
dryajov: > > If we decide to copy it's just a few lines of code worth of changes, I don't think this is a major issue and we can try both ways. For cancellations, I think we have a good way of handling them in the code right now, we just have to make sure we've covered all the code paths.
> 
> Okay if we cover all the cancellation pathways that'd be fine. For me it's an important building block and getting the rest of the pieces can be hard enough by themselves. It's why I focus on them. I'll read a bit more closely what you did for the cancellation pieces.

I definitely agree 👍

datastore/threads/databuffer.nim
```diff
@@ -50,15 +51,17 @@ proc new*[T: byte | char](tp: type DataBuffer, data: openArray[T]): DataBuffer =
...
+    copyMem(result[].buf, baseAddr data, data.len)
 
 converter toSeq*(self: DataBuffer): seq[byte] =
   ## convert buffer to a seq type using copy and either a byte or char
   ##
 
   result = newSeq[byte](self.len)
   if self.len() > 0:
-    copyMem(addr result[0], unsafeAddr self[].buf[0], self.len)
+    copyMem(addr result[0], addr self[].buf[0], self.len)
```
elcritch, line 64: Ah good! I was thinking about that this weekend.

dryajov: yeah, but maybe this should also be switched to `baseAddr`? Not sure.

dryajov: Or the other way around, switch `baseAddr` to `addr`?

dryajov: Imo, `addr` is probably safer, `baseAddr` does a straight up `cast`.

dryajov:
This https://github.com/codex-storage/nim-codex/pull/552 works now.

PR #49, opened 2023-09-18
Threadpool semaphore ref ctx type

elcritch:
Experimenting with using TaskCtx flags and a `GC_ref` to handle corner case during cancellation. 

However, it seems to pull up some other issues with `Datastore` being a GC type and using it on multiple threads.

### DIFF
--- a/datastore/threads/threadproxyds.nim
+++ b/datastore/threads/threadproxyds.nim
@@ -35,12 +35,12 @@ type
   ThreadResult[T: ThreadTypes] = Result[T, DataBuffer]
 
   TaskCtx[T: ThreadTypes] = ref object
-    ds: ptr Datastore
-    res: ThreadResult[T]
+    ds*: ptr Datastore
+    res*: ThreadResult[T]
     cancelled: Atomic[bool]
     isActive: Atomic[bool]
     semaphore: AsyncSemaphore
-    signal: ThreadSignalPtr
+    signal*: ThreadSignalPtr
 
   ThreadDatastore* = ref object of Datastore
     tp: Taskpool
--- a/datastore/threads/threadproxyds.nim
+++ b/datastore/threads/threadproxyds.nim
@@ -62,8 +62,8 @@ proc new*[T](
     ds: Datastore,
 ): ref TaskCtx[T] =
   result = (ref TaskCtx[T])()
-  result.ds = unsafeAddr(ds) ## 
-    ## doing this appears to break. previously it was using `addr(ds)`
+  result.ds = unsafeAddr(ds) ##\
+    ## doing this appears to break things. previously it was using `addr(ds)`
     ## and reverting to those lets the tests get further. 
     ## 
     ## however, that seems to mean that `addr(ds)` we're taking the 
--- a/datastore/threads/threadproxyds.nim
+++ b/datastore/threads/threadproxyds.nim
@@ -74,6 +74,9 @@ proc new*[T](
     ## of the `ds` argument, which is a temporary stack location.
     ## 
     ## not sure how to fix this while using GC types
+    ## just taking the address of the var location sorta works, but
+    ## crashes now as well, but later. likely due to datastore now being
+    ## on two GC heaps?
 
   echo ""
   echo "TaskCtx:new: ", "addrOf: ", addrOf(result).pointer.repr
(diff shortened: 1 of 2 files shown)


datastore/threads/threadproxyds.nim
```diff
@@ -51,9 +53,43 @@ type
...
+    ## however, that seems to mean that `addr(ds)` we're taking the 
+    ## address of the `var ds: Datastore` location, and not the actual
+    ## Datastore:ObjectType. As in `ds: ptr(ref Datastore:ObjectType)`. 
+    ## 
+    ## so doing the `unsafeAddr` would mean we're taking the ptr location
+    ## of the `ds` argument, which is a temporary stack location.
+    ## 
+    ## not sure how to fix this while using GC types
+    ## just taking the address of the var location sorta works, but
+    ## crashes now as well, but later. likely due to datastore now being
```
elcritch, line 78: @dryajov related to the question of addr vs baseAddr. I've tried a few variations. But the `ptr Datastore` keeps segfaulting at some point. The `ptr TaskCtx` change I did appears fine and it's contents are what they're set to be.

elcritch: Here's a simplified version: https://github.com/codex-storage/nim-datastore/pull/50 that does the same thing. If you can capture the `var Datastore` address it sorta works, but I think pointers to a bigger issue with the GC type. Doing stress tests last week I saw lots of GC issues with other ref types.

datastore/threads/threadproxyds.nim
```diff
@@ -31,15 +31,17 @@ type
...
+  ThreadTypes* = void | bool | SomeInteger | DataBuffer | tuple | Atomic
+  ThreadResult*[T: ThreadTypes] = Result[T, DataBuffer]
+
+  TaskCtx*[T: ThreadTypes] = object
+    head: int
+    ds*: ptr Datastore
+    res*: ThreadResult[T]
+    cancelled: Atomic[bool]
+    isActive: Atomic[bool]
     semaphore: AsyncSemaphore
```
dryajov, line 43: Why would you want to do this?

datastore/threads/threadproxyds.nim
```diff
@@ -51,9 +53,43 @@ type
...
+  result = cast[ptr TaskCtx[T]](ctx)
+  echo "ADDR_OF: ", result.pointer.repr
+
+proc new*[T](
+    ctx: typedesc[TaskCtx[T]],
+    ds: Datastore,
+): ref TaskCtx[T] =
+  result = (ref TaskCtx[T])()
+  result.ds = unsafeAddr(ds) ##\
+    ## doing this appears to break things. previously it was using `addr(ds)`
```
dryajov, line 66: You really don't want to do this, because there is no way for the compiler to trace the lifetime of `ds` ref type, that is why you cant use `addr` in this context and you are forced to break out of the compiler's safety with `unsafeAddr`. 

On top of that, you're turning `TaskCtx` into a `ref` for a reason, because we actually want to control the lifetime of the `ctx` and constrain it to the stack's lifetime.

datastore/threads/threadproxyds.nim
elcritch: To keep the result and other args as valid memory in the event that a cancellation occurs and the thread-task is using the memory. To do that you need to ensure the lifetime of the `TaskCtx` is for as long as the anyone who is using this object. 

As I mentioned in my comment, instead of turning `TaskCtx` into a ref, you could keep it in the Future's. In that event to do it safely you'd need to do a spinlock or similar on the `isActive` until the task-thread can be cancelled or finished.

datastore/threads/threadproxyds.nim
elcritch: > You really don't want to do this, because there is no way for the compiler to trace the lifetime of ds ref type, that is why you cant use addr in this context and you are forced to break out of the compiler's safety with unsafeAddr.

Yes, it's not a good way to approach it, but it highlights the challenge here. I think just changing it the type of `ds: ptr Datastore` to `ds: Datastore` would be better. However, that also breaks the GCs as I've been finding in my testing.

That lifetime analysis isn't quite right as in general it's not safe to 
use `addr` here either. It's just *slightly* more safe since there's a guaranteed stack address. The compiler can't determine the lifetime or use it to track the datastore's life when you use `addr` anymore than with `unsafeAddr`. They both become `ptr Datastore` which the GC doesn't understand. It sorta works because the `var ds` here get's captured into the Future whose lifetime generally exceeds the task, but not always.

At least the Future's lifetime exceeds it in the normal control flow, but passing it to the thread-tasks still bypasses that. The situation is similar to if you try to capture a `var ds` into a closure - the compiler will error because doing so isn't safe even in a non-ref object. So just using a `ds: Datastore` and then an additional `GC_ref` would properly increase the ref count of the `ds` which could then be decreased when both sides are done.

PR #58, opened 2023-11-08
 Concurrent datastore interface & sqlite implementation

tbekas:
Interface and sqlite implementation for concurrently safe modifications of data.

Currently, to modify a data it's required to do a sequence of `get-modify-put` operations, which can lead to erroneous results when there are other concurrent `get-modify-put` operations.


datastore/sql/sqliteds.nim
```diff
@@ -29,6 +29,99 @@ proc `readOnly=`*(self: SQLiteDatastore): bool
...
+    var maybeNewData: ?seq[byte]
+
+    try:
+      maybeNewData = await fn(maybeCurrentData)
+    except CatchableError as err:
+      return failure("Error running modify function: " & err.msg)
+
+    if maybeCurrentData == maybeNewData:
+      # no need to change any stored value
+      break;
```
dryajov: why the `;`?

datastore/sql/sqliteds.nim
```diff
@@ -29,6 +29,99 @@ proc `readOnly=`*(self: SQLiteDatastore): bool
 proc timestamp*(t = epochTime()): int64 =
   (t * 1_000_000).int64
 
+const initVersion* = 0.int64
+
+method modify*(self: SQLiteDatastore, key: Key, fn: ModifyAsync): Future[?!void] {.async.} =
```
dryajov: Shouldn't this be directly used by `put`, instead of being part of the public interface?

datastore/sql/sqliteds.nim
tbekas: Old habit, fixing it :)

datastore/sql/sqliteds.nim
tbekas: Not sure if I understand this comment, you mean that the `put` would call `modify` underneath? 

If so, then it's not necessary, because `put` is an unconditional write. With `put` you write data, regardless of whatever was in the store prior to the write operation. We _could_ use `modify` for it, but we would get no benefit out of doing so, since the whole promise of `modify` is to support conditional writes (depending on what's currently in the store). Such version of `put` would be actually a bit slower than the current version because in `modify` we access db at least two times (read & write), while `put` currently does it exactly once (just write).

dryajov: Yes, the problem is that we're now creating separate datastore interfaces for different applications, this is not how this was designed and it would break all the composability benefits that the current design brings . 

I think here we have two options:

- Add the modify methods to the top level interface
- Move the modify into `put` and use a flag to `force` insertion

I prefer the later, because it would keep the interface consistent with the remaining stores.

datastore/concurrentds.nim
```diff
@@ -0,0 +1,50 @@
...
+export key, query, types, datastore
+
+push: {.upraises: [].}
+
+type
+  Function*[T, U] = proc(value: T): U {.upraises: [], gcsafe, closure.}
+  Modify* = Function[?seq[byte], ?seq[byte]]
+  ModifyAsync* = Function[?seq[byte], Future[?seq[byte]]]
+
+method modify*(self: ConcurrentDatastore, key: Key, fn: Modify): Future[?!void] {.base, locks: "unknown".} =
```
dryajov: If you have a sync callback, there is no reason to make the top level method `async`, i.e. return a `Future`.

datastore/sql/sqliteds.nim
dryajov: Maybe a third option is to add an overload for `put` that takes the `modify` callback, in which case it would only put if a) no value is present or b) if the value is present it succeeds with updating it.

datastore/concurrentds.nim
tbekas: This method from `Datastore` interface doesn't have any callback at all and it returns future:

`method put*(self: Datastore, key: Key, data: seq[byte]): Future[?!void]`

datastore/sql/sqliteds.nim
tbekas: >Yes, the problem is that we're now creating separate datastore interfaces for different applications, this is not how this was designed and it would break all the composability benefits that the current design brings .

How that is breaking composability? If something was composing before this change, it will continue doing so after. `SQLiteDatastore` continues to implement the `Datastore` interface transitively through the `ConcurrentDatastore`. This is not a breaking change.

>Add the modify methods to the top level interface

Ok, but then I would have to implement it on all stores, some of these implementations may be challenging and for now they would simply remain unused. I find it way more reasonable to have `modify` methods in a separate interface and possibly migrate them to main `Datastore` interface once we have them implemented for all stores (`FSDatastore`, `MountedDatastore`, `TieredDatastore` and `SQLiteDatastore`).

>Move the modify into put and use a flag to force insertion

Same issue as above, `put` is implemented by all stores, so that would imply adding implementation for all of them.

>add an overload for put that takes the modify callback

This is pretty much the same as above.

datastore/concurrentds.nim
dryajov: True, and this is one thing I want to change, there should be a sync and an async interface, (but this ins't as straightforward), in which case those methods won't return a `Future`. 

The problem with what you're doing is that you are adding a blocking callback to a non-blocking method call, when executing the callback, it's going to block the "non-blocking" call.

tbekas: Let's be specific. By blocking callback you mean a function containing call that will block a thread? Like [std/os sleep](https://nim-lang.org/docs/os.html#sleep%2Cint) for example? If so, what prevents users from using such call in an async function?

dryajov: Nothing, but why return a `Future` at all if the callback signature is already _blocking_ and there is no way of calling it in a non blocking way - you see the discrepancy? 

All I'm  saying is that, if you "promising" that a method/proc is asynchronous, it doesn't make sense to pass in a synchronous primitive, so either make the callback async or remove the `Future` return type. You already have an async version of this method, so the logical step would be to have a sync version of it, simply do not return a `Future` from the sync version.

tbekas: >the callback signature is already blocking

A signature cannot be blocking. An implementation can be blocking. There's nothing in this signature that says that the implementation has to be blocking. It can be a simple pure function (and that's the intention). Example: adding +1 to a number (like increasing a ref count) - this is a pure and non-blocking operation.

The method signature returns Future because a store that will implement it will likely do an IO operation underneath, similarly as `put` is doing.

dryajov: > A signature cannot be blocking. An implementation can be blocking.

A signature should communicate the intent of the underlying implementation. In this context, the signature is inconsistent - it returns a future but executes a synchronous callback. 

> There's nothing in this signature that says that the implementation has to be blocking.

There is no way to execute this callback asynchronously, because there is no way of communicating a future result or continuation, so by definition, this callback will always block the main event loop.

> It can be a simple pure function (and that's the intention). 

I understand, but again, there is no way to tell that from the API, anyone can simply assume that since it still returns a `Future`, it would be safe to do blocking or long running operations in the callback. This is purely about a) being consistent and b) communicating the correct semantics to the user.

You have to realize, that just because you're wrapping a sync operation in a future, it wont magically become an async operation. All tasks share the same thread, so having a sync proc return a `Future` is pointless.

--------

In general, we should have two separate versions of the api, one that is sync and another one that is async, this is however not straightforward to do. 

The easiest approach is to provide two version of the apis, one async and one explicitly sync, however thats a bit ugly and worst, it usually leads to code duplication.

In `asyncdispatch` (the std version of chronos), there is a handy macro called `multisync` that does this automatically, i.e. it generates two separate versions of a proc/method, one sync and one async, you can look at how it's implemented in the http client - https://nim-lang.org/docs/httpclient.html. We don't have this in chronos, however it should be easy to add.

datastore/sql/sqliteds.nim
```diff
@@ -29,6 +29,122 @@ proc `readOnly=`*(self: SQLiteDatastore): bool
...
+      if err =? self.db.rollbackStmt.exec().errorOption:
+        return failure(err)
+      return failure("Unexpected number of changes, expected either 0 or 1, was " & $changes)
+
+  if retriesLeft == 0:
+    return failure("Retry limit exceeded")
+
+  return success(aux)
+
+method modifyGet*(self: SQLiteDatastore, key: Key, fn: ModifyGet): Future[?!seq[byte]] {.async.} =
```
dryajov: Please remove `Future` and `async` from this method, even better, simply remove this method altogether for now and just leave the async version. As mentioned here https://github.com/codex-storage/nim-datastore/pull/58#discussion_r1406676393, a sync interface is planned, but it isn't straighforward to add, so for now the best corse of action is to stick with the async interface.

datastore/sql/sqliteds.nim
```diff
@@ -29,6 +29,122 @@ proc `readOnly=`*(self: SQLiteDatastore): bool
...
+    let res = await fn(maybeValue)
+    let ignoredAux = newSeq[byte]()
+    return (res, ignoredAux)
+
+  if err =? (await self.modifyGet(key, wrappedFn)).errorOption:
+    return failure(err)
+  else:
+    return success()
+
+method modify*(self: SQLiteDatastore, key: Key, fn: Modify): Future[?!void] {.async.} =
```
dryajov: Same as above, please drop this for now.

datastore/concurrentds.nim
tbekas: >a blocking callback

>a synchronous callback

You're using term `synchronous` and `blocking` interchangeably, as if they would be the same thing - which is not the case.

This call to `+` is synchronous and non blocking `let x = 1 + 1`.
This call to `sleep` is synchronous and blocking `sleep(100)`.

>anyone can simply assume that since it still returns a Future, it would be safe to do blocking or long running operations in the callback.

It's never-ever safe to do any blocking or long running calls in the code if this code is going to be be executed on a main compute thread (the event loop thread). And that's it. If anyone assumes that it's safe to use a blocking call in the callback, then well - that's simply a wrong assumption.

Anyway, if you're still unconvinced I'm going to remove these methods, just to get this review passed.

dryajov: `synchronous` and `blocking` are interchangeable in this context, both long running computations and blocking IO will have the same effect on the event loop.

> If anyone assumes that it's safe to use a blocking call in the callback, then well - that's simply a wrong assumption.

That isn't the case tho, any long running computation can be broken down into smaller pieces, yielding control back to the main event loop, for example you can use `idleAsync` or `sleepAsync`, but this is only possible in an async callback, i.e. one that returns a continuation.

dryajov: But I think the finer point of my argument is being missed thought the entire exchange - it is simply wrong to mix incompatible execution policies, this brings confusion and ultimately leads to a bad API.

dryajov (approved):
LGTM after the last comments re sync methods have been addressed.

datastore/concurrentds.nim
tbekas: Ok, I understand. Methods accepting non-async callbacks got removed in latest changes.

datastore/sql/sqliteds.nim
tbekas: Implemented `modify` for all stores.

.github/workflows/tests.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         cache_nonce: [ 1 ]
-        nim_version: [ 1.6.14 ]
+        nim_version: [ 1.6.16 ]
```
dryajov: 16 is buggy, I think 18 is out now?

dryajov: https://nim-lang.org/blog/2023/12/19/versions-1618-202-released.html

datastore/datastore.nim
```diff
@@ -12,6 +13,9 @@ push: {.upraises: [].}
 
 type
   BatchEntry* = tuple[key: Key, data: seq[byte]]
+  Function*[T, U] = proc(value: T): U {.upraises: [CatchableError], gcsafe, closure.}
```
dryajov: We shouldn't use `upraises` anymore, just use the bare `raises: []`, upraises was a compatibility workaround for versions prior to 1.6.14, which didn't support `raises: []` and always needed an exception such as `raises: [Defect]`.

Maybe can be done in a followup cleanup, if there are more upraises used elsewhere in the stores (can't remember now).

datastore/fsds.nim
```diff
@@ -217,6 +219,18 @@ method query*(
   iter.next = next
   return success iter
 
+method modifyGet*(
+  self: FSDatastore,
+  key: Key,
+  fn: ModifyGet): Future[?!seq[byte]] =
+  defaultModifyGetImpl(self, self.lock, key, fn)
```
dryajov: Hmm... as far as I can tell, this will lock the entire datastore regardless of which key is being accessed. I've been thinking about locking for the stores for some time and I think the best approach would be to have a lock per key.

One way of doing this would be to have a table with keys and use the `mgetOrPut` method, like so:

```nim
   try:
     let lock = self.locks.mgetOrPut(key, newAsyncLock())
     defaultModifyGetImpl(self, self.lock, key, fn)
   finally:
     if not lock.locked:
        self.locks.del(key)
```

But thinking about this some more, perhaps, we don't want to handle locking inside the `modifyGet` itself and leave it to the store, since it has a lot more context to make the appropriate decisions, some stores don't even require locking, for example sqlite handles locking internally.

dryajov (changes requested):
Awesome job! I really like how this is turning out, really clean and concise code 👍 

My only request is that we make the change with the locking as per my comment. I'd suggest we leave it out of the default modifyGet implementation and make it per key, otherwise it's a foot gun waiting to go off.

.github/workflows/tests.yml
tbekas: Thanks for the info, fixing it

datastore/datastore.nim
tbekas: Thanks for the info. I simply changed it to `raises: [CatchableError]` and the code is compiling and tests are passing, so I think it's working as intended.

datastore/fsds.nim
tbekas: I fully agree that locking per key is better than locking per entire ds, I think I oversimplified the default implementation. I used your code and it's working well and tests are passing.

Regarding the comment

>But thinking about this some more, perhaps, we don't want to handle locking inside the `modifyGet` itself and leave it to the store

It is already the case, `modify` and `modifyGet` are just interface methods. Proc `defaultModifyGetImpl` is just some default implementation that can be used or not.

dryajov (approved):
LGTM

PR #59, opened 2023-12-15
Updates for Chronos V4

gmega:
This PR adds minor exception effect annotations so that datastore is compatible with Chronos V4. It also explicitly tracks chronos#head in the nimble file until Chronos 4.0.0 is officially out as we'll otherwise get misleading CI results.

### DIFF
--- a/datastore.nimble
+++ b/datastore.nimble
@@ -8,7 +8,7 @@ license       = "Apache License 2.0 or MIT"
 
 requires "nim >= 1.2.0",
          "asynctest >= 0.4.3 & < 0.5.0",
-         "chronos#head", # FIXME change to Chronos >= 4.0.0 once it's out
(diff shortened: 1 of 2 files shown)


datastore.nimble
```diff
@@ -7,8 +7,8 @@ description   = "Simple, unified API for multiple data stores"
 license       = "Apache License 2.0 or MIT"
 
 requires "nim >= 1.2.0",
-         "asynctest >= 0.3.1 & < 0.4.0",
-         "chronos",
+         "asynctest >= 0.4.3 & < 0.5.0",
+         "chronos#head", # FIXME change to Chronos >= 4.0.0 once it's out
```
dryajov: can we pin this to the specific head commit for now?

dryajov (approved):
LGTM, but lets pin to the current head commit.

PR #60, opened 2024-01-13
Proper dispose

dryajov:
Fixes `SIGSIGV` on db close, due to not disposing of statments and connections properly.

### DIFF
--- a/datastore/key/namespace.nim
+++ b/datastore/key/namespace.nim
@@ -1,6 +1,5 @@
 
 import std/hashes
-import std/strformat
 
 import pkg/questionable
 import pkg/questionable/results
@@ -19,32 +18,26 @@ type
 
 func init*(T: type Namespace, field, value: string): ?!T =
   if value.contains(Delimiter):
-      return failure (&"value string must not contain Delimiter '{Delimiter}'")
-        .catch.expect("should not fail")
+      return failure ("value string must not contain Delimiter " & Delimiter)
 
   if value.contains(Separator):
-    return failure (&"value string must not contain Separator {Separator}")
-      .catch.expect("should not fail")
+    return failure ("value string must not contain Separator " & Separator)
 
   if field.contains(Delimiter):
-    return failure (&"field string must not contain Delimiter {Delimiter}")
-      .catch.expect("should not fail")
+    return failure ("field string must not contain Delimiter " & Delimiter)
 
   if field.contains(Separator):
-    return failure (&"field string must not contain Separator {Separator}")
-      .catch.expect("should not fail")
+    return failure ("field string must not contain Separator " & Separator)
 
   success T(field: field, value: value)
 
 func init*(T: type Namespace, id: string): ?!T =
   if id.len > 0:
     if id.contains(Separator):
-      return failure (&"id string must not contain Separator {Separator}")
-        .catch.expect("should not fail")
+      return failure (&"id string must not contain Separator " & Separator)
 
     if id.count(Delimiter) > 1:
-      return failure (&"id string must not contain more than one {Delimiter}")
-        .catch.expect("should not fail")
+      return failure (&"id string must not contain more than one " & Delimiter)
 
   let
     (field, value) = block:
--- a/datastore/sql/sqliteds.nim
+++ b/datastore/sql/sqliteds.nim
@@ -228,8 +228,7 @@ method query*(
 
   let
     queryStmt = QueryStmt.prepare(
-      self.db.env, queryStr).expect("should not fail")
-
+      self.db.env, queryStr).expect("Query prepare should not fail")
     s = RawStmtPtr(queryStmt)
 
   var
@@ -263,7 +262,7 @@ method query*(
       let
         key = Key.init(
           $sqlite3_column_text_not_null(s, QueryStmtIdCol))
-          .expect("should not fail")
+          .expect("Key should should not fail")
 
         blob: ?pointer =
           if query.value:
@@ -308,7 +307,7 @@ method query*(
   iter.dispose = proc(): Future[?!void] {.async.} =
     discard sqlite3_reset(s)
     discard sqlite3_clear_bindings(s)
-    s.dispose
+    iter.next = nil
     return success()
 
   iter.next = next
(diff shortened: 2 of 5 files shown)


datastore/sql/sqlitedsdb.nim
```diff
@@ -279,25 +279,51 @@ proc getDBFilePath*(path: string): ?!string =
...
-  self.getStmt.dispose
-  self.beginStmt.dispose
-  self.endStmt.dispose
-  self.rollbackStmt.dispose
-  self.getVersionedStmt.dispose
-  self.updateVersionedStmt.dispose
-  self.insertVersionedStmt.dispose
-  self.deleteVersionedStmt.dispose
-  self.getChangesStmt.dispose
+proc close*(self: var SQLiteDsDb) =
```
elcritch, line 282: It doesn't look like `self` needs to be a var, but it does seem to indicate intent a bit better.

datastore/sql/sqliteds.nim
```diff
@@ -308,7 +307,7 @@ method query*(
   iter.dispose = proc(): Future[?!void] {.async.} =
     discard sqlite3_reset(s)
     discard sqlite3_clear_bindings(s)
-    s.dispose
+    iter.next = nil
```
elcritch, line 310: I think we'll still need to do a `dispose` on the query statement. Otherwise it looks like it'll leak the query statements. 

If this dispose is causing the issue then possibly we're calling the iterator's dispose multiple times? I've been meaning to create a PR to pull out some of the fixes I did in the threaded datastore for how codex uses the query iterator. Although most of those issues were due to Codex not calling dispose on the iterator, IIRC. 

Possibly we might need to add `isFinished` flag to ensure we only call dispose once? Alternatively a destructor for QueryIter could work.

datastore/sql/sqlitedsdb.nim
elcritch: Ah nevermind, I see the nil setter down below.

datastore/sql/sqliteutils.nim
```diff
@@ -250,7 +257,7 @@ proc query*(
   query: string,
   onData: DataProc): ?!bool =
 
-  let
+  var
```
tbekas, line 260: Do they need to be vars?

datastore/key/namespace.nim
```diff
@@ -19,32 +18,26 @@ type
...
 
   if field.contains(Delimiter):
-    return failure (&"field string must not contain Delimiter {Delimiter}")
-      .catch.expect("should not fail")
+    return failure ("field string must not contain Delimiter " & Delimiter)
 
   if field.contains(Separator):
-    return failure (&"field string must not contain Separator {Separator}")
-      .catch.expect("should not fail")
+    return failure ("field string must not contain Separator " & Separator)
```
tbekas, line 30: Just out of curiosity, why did you decide to replace string interpolation with concatenation?

tbekas (approved):
LGTM

datastore/sql/sqliteds.nim
dryajov: > I think we'll still need to do a dispose on the query statement. Otherwise it looks like it'll leak the query statements.

`dispose` just calls `sqlite3_reset(s)` and sets the statement to `nil`, it's not required here because we're already calling `sqlite3_reset` on line 308.

datastore/sql/sqliteutils.nim
dryajov: Yes, unfortunately, we need to set the cached statements to `nil` on close/destroy, which requires `SQLiteDsDb` to be var.

datastore/key/namespace.nim
dryajov: `&`/`{}` raise `ValueError`, which is why there was the weird `catch.expect` attached there, which just feels cumbersome for very little gain, IMO...

dryajov:
Another approach would be to use `AutoDisposed` and `disposeIfUnreleased` instead of adding an explicit `nil` to dispose -  `disposeIfUnreleased` does that already:

```nim
proc disposeIfUnreleased*[T](x: var AutoDisposed[T]) =
  mixin dispose
  if x.val != nil: dispose(x.release)
```

But it also feels a bit cumbersome for very little gain?

This api definitely needs some cleanup/love, but at least this fixes the segfaults and invalid heap/stack access that I was getting from the sanitizer.

dryajov:
I'll merge this for now, but we should revisit this soon.

datastore/key/namespace.nim
tbekas: Thanks for the explanation 👌

datastore/sql/sqliteds.nim
elcritch: It looks like here `dispose` for the `RawStmtPtr` is calling `sqlite3_finalize` function not `sqlite3_reset`. 

https://github.com/codex-storage/nim-datastore/blob/d02f2e7380410137fd2cdc9bd0d06c0004f7e8ff/datastore/sql/sqliteutils.nim#L92

Also the sqlite3 [docs](https://www.sqlite.org/c3ref/finalize.html) say `sqlite3_reset` is just to reset the query state, not to clean it up:

> The sqlite3_finalize(S) routine can be called at any point during the life cycle of [prepared statement](https://www.sqlite.org/c3ref/stmt.html) S: before statement S is ever evaluated, after one or more calls to [sqlite3_reset()](https://www.sqlite.org/c3ref/reset.html), or after any call to [sqlite3_step()](https://www.sqlite.org/c3ref/step.html) regardless of whether or not the statement has completed execution.

elcritch: @dryajov just pinging you to make sure you saw my comment above.

dryajov: > It looks like here `dispose` for the `RawStmtPtr` is calling `sqlite3_finalize` function not `sqlite3_reset`.
> 
> https://github.com/codex-storage/nim-datastore/blob/d02f2e7380410137fd2cdc9bd0d06c0004f7e8ff/datastore/sql/sqliteutils.nim#L92
> 
> Also the sqlite3 [docs](https://www.sqlite.org/c3ref/finalize.html) say `sqlite3_reset` is just to reset the query state, not to clean it up:
> 
> > The sqlite3_finalize(S) routine can be called at any point during the life cycle of [prepared statement](https://www.sqlite.org/c3ref/stmt.html) S: before statement S is ever evaluated, after one or more calls to [sqlite3_reset()](https://www.sqlite.org/c3ref/reset.html), or after any call to [sqlite3_step()](https://www.sqlite.org/c3ref/step.html) regardless of whether or not the statement has completed execution.

Ah, you might be right, I'll take another look.

PR #62, opened 2024-05-08
TypedDatastore implementation

tbekas:
Wrapper for Datastore with basic functionality of automatically converting stored values from some user defined type `T` to `seq[byte]` and vice-versa.

### DIFF
--- a/datastore/typedds.nim
+++ b/datastore/typedds.nim
@@ -1,0 +1,180 @@
+import std/options
+import std/macros
+
+import pkg/questionable
+import pkg/questionable/results
+import pkg/chronos
+import pkg/chronos/futures
+
+import ./datastore
+
+## Wrapper for Datastore with basic functionality of automatically converting
+## stored values from some user defined type `T` to `seq[byte]` and vice-versa.
+##
+## To use this API you need to provide decoder and encoder procs.
+##
+## Basic usage
+## ==================
+## .. code-block:: Nim
+##   import pkg/stew/byteutils
+##   import pkg/questionable
+##
+##   let
+##     tds = ds.asTypedDs()
+##     key = Key.init("p").tryGet()
+##
+##   type Person = object
+##     name: string
+##   proc encode(t: Person): seq[byte] =
+##     t.name.toBytes()
+##   proc decode(T: type Person, bytes: seq[byte]): ?!T =
+##     success(Person(name: string.fromBytes(bytes)))
+##
+##   discard await tds.put(key, Person(name: "John"))
+##
+##   let result = (await get[Person](tds, key)).tryGet()
+##
+##   assert:
+##     result == Person(name: "John")
+
+
+type
+  TypedDatastore* = ref object of RootObj
+    ds*: Datastore
+
+  Modify*[T] = proc(v: ?T): Future[?T] {.raises: [CatchableError], gcsafe, closure.}
+  ModifyGet*[T, U] = proc(v: ?T): Future[(?T, U)] {.raises: [CatchableError], gcsafe, closure.}
+
+  QueryResponse*[T] = tuple[key: ?Key, value: ?!T]
+  GetNext*[T] = proc(): Future[?!QueryResponse[T]] {.raises: [], gcsafe, closure.}
+  QueryIter*[T] = ref object
+    finished*: bool
+    next*: GetNext[T]
+    dispose*: IterDispose
+
+# Helpers
+template requireDecoder*(T: typedesc): untyped =
+  when (compiles do:
+    let _: ?!T = T.decode(newSeq[byte]())):
+    discard
+  else:
+    {.error: "provide a decoder: `proc decode(T: type " & $T & ", bytes: seq[byte]): ?!T`".}
+
+template requireEncoder*(T: typedesc): untyped =
+  when (compiles do:
+    let _: seq[byte] = encode(default(T))):
+    discard
+  else:
+    {.error: "provide an encoder: `proc encode(a: " & $T & "): seq[byte]`".}
+
+# Original Datastore API
+proc has*(self: TypedDatastore, key: Key): Future[?!bool] {.async.} =
+  await self.ds.has(key)
+
+proc contains*(self: TypedDatastore, key: Key): Future[bool] {.async.} =
+  return (await self.ds.has(key)) |? false
+
+proc delete*(self: TypedDatastore, key: Key): Future[?!void] {.async.} =
+  await self.ds.delete(key)
+
+proc delete*(self: TypedDatastore, keys: seq[Key]): Future[?!void] {.async.} =
+  await self.ds.delete(keys)
+
+proc close*(self: TypedDatastore): Future[?!void] {.async.} =
+  await self.ds.close()
+
+# Conversion Datastore -> TypeDatastore
+proc asTypedDs*(ds: Datastore): TypedDatastore =
+  TypedDatastore(ds: ds)
+
+# TypedDatastore API
+proc put*[T](self: TypedDatastore, key: Key, t: T): Future[?!void] {.async.} =
+  requireEncoder(T)
+
+  await self.ds.put(key, t.encode)
+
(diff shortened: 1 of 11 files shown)


tests/datastore/modifycommontests.nim
```diff
@@ -108,17 +108,10 @@ proc modifyTests*(
...
-      check:
-        res.errorOption.map((err) => err.msg) == none(string)
-      for b in res |? @[]:
-        check:
-          b == 123.byte
-    else:
-      check:
-        res == success(@[byte 123])
+    check:
+      aux == (123.byte).repeat(dsCount)
```
benbierens, line 114: I don't understand the changes in this file at all. Please explain.

tests/datastore/sql/testsqlitedsdb.nim
```diff
@@ -7,6 +7,7 @@ import pkg/stew/byteutils
 import pkg/sqlite3_abi
 import pkg/datastore/key
 import pkg/datastore/sql/sqlitedsdb
+import pkg/datastore/sql/sqliteutils
```
benbierens, line 10: an import was added but nothing else in this test file changes. Is it needed?

tests/datastore/typeddscommontests.nim
```diff
@@ -0,0 +1,140 @@
...
+    proc returningAux(maybeI: ?int): Future[(?int, string)] {.async.} =
+      if i =? maybeI:
+        (some(i + 1), "foo")
+      else:
+        (int.none, "bar")
+
+    (await tds.put(key, 44)).tryGet()
+
+    check:
+      "foo".repeat(dsCount) == (await tds.modifyGet(key, returningAux)).tryGet()
```
benbierens, line 93: why the .repeat(dsCount) ?
does the modifyGet not always return "foo" once?

tests/datastore/typeddscommontests.nim
```diff
@@ -0,0 +1,140 @@
...
+        "c": 33,
+        "d": 44
+      }.toTable
+      Root = Key.init("/querytest").tryGet()
+
+    for k, v in source:
+      let key = (Root / k).tryGet()
+      (await tds.put(key, v)).tryGet()
+
+    let iter = (await query[int](tds, Query.init(Root))).tryGet()
```
benbierens, line 121: I think it's possible with a query to return multiple 'nested keys'.
If you query "/a", you can receive ["/a", "/a/b", "/a/c"] for example.
Maybe it's a good idea to add a testcase for this sort of behavior.

benbierens (commented):
I think the TypedDatastore is a good idea.

tests/datastore/modifycommontests.nim
tbekas: So there's a case of `TieredDatastore` that need this dedicated handling here. `TieredDatastore` is defined as a composition of few other stores.
```
TieredDatastore* = ref object of Datastore
    stores: seq[Datastore]
```

When calling methods like `put` or `modify`, all stores gets updated.

API has also another method called `modifyGet` that not only update the value but also return some auxiliary value (could be anything that developer defines like for example previous value before update).

When we have multiple stores, we have multiple auxiliary values and they have to be combined somehow. The arbitrary choice I've made some time ago is to simply concatenate these values.

In test above the single auxiliary value we return is just `@[byte 123]` (a sequence with a single byte). When we concatenate multiple auxiliary values we get `@[byte 123, 123, 123, ...]`, which is the same as `(123.byte).repeat(n)`. `n` is just number of stores we use (which is 2 in this case).

I've decided to make this test consistent with what's in the `typedscommontests.nim` and use `dsCount` instead of `multiAux` parameter, however the previous version would also work.

tests/datastore/sql/testsqlitedsdb.nim
tbekas: Yes, because `sqliteutils` is no loger exported by `sqlitedsdb` (one import above)

tests/datastore/typeddscommontests.nim
tbekas: Yes, the `TieredDatastore` case. I've explained it in [this comment](https://github.com/codex-storage/nim-datastore/pull/62/files#r1593983836)

tests/datastore/typeddscommontests.nim
tbekas: The detailed cases of how querying works is already covered in `querycommontests.nim` module, there's [this](https://github.com/codex-storage/nim-datastore/blob/d02f2e7380410137fd2cdc9bd0d06c0004f7e8ff/tests/datastore/querycommontests.nim#L29) test case that covers that exact scenario.

The goal of this test is to check if the conversion between `seq[byte]` and `T` works well within `query` method.

datastore/typedds.nim
```diff
@@ -0,0 +1,180 @@
...
+  QueryResponse*[T] = tuple[key: ?Key, value: ?!T]
+  GetNext*[T] = proc(): Future[?!QueryResponse[T]] {.raises: [], gcsafe, closure.}
+  QueryIter*[T] = ref object
+    finished*: bool
+    next*: GetNext[T]
+    dispose*: IterDispose
+
+# Helpers
+template requireDecoder*(T: typedesc): untyped =
+  when (compiles do:
```
dryajov: nitpick - just use `not` instead of the `discard/else`

datastore/typedds.nim
```diff
@@ -0,0 +1,180 @@
...
+  await self.ds.delete(key)
+
+proc delete*(self: TypedDatastore, keys: seq[Key]): Future[?!void] {.async.} =
+  await self.ds.delete(keys)
+
+proc close*(self: TypedDatastore): Future[?!void] {.async.} =
+  await self.ds.close()
+
+# Conversion Datastore -> TypeDatastore
+proc asTypedDs*(ds: Datastore): TypedDatastore =
```
dryajov: Not sure about this, not a very nim'ish approach, `init` should do just fine here? If you want automatic conversions between types you can also look into `converter`s

dryajov (commented):
Overall looks good, however, I'm not sure about the `requireDecoder/Encoder` methods. Nim has a mechanism for this already https://nim-lang.org/docs/manual.html#generics-mixin-statement, which should allow binding to symbols at the instantiation point, in other words, if you have an object that defines `encode/decode` methods, and you declare `mixin encode` and `mixin decode` in the respective `get/put/etc...` methods, they should be picked up at the instantiation point. Now, mixin isn't without quirks, but it's worth trying it out before rolling our own mechanism.

datastore/typedds.nim
tbekas: Makes sense!

datastore/typedds.nim
tbekas: I like the approach with the `init` function. Fixing it

tbekas:
@dryajov I checked the `mixin` functionality and I got the same result using `mixin` as not using anything. We could go with not using anything (so no `mixin` and no `requireDecoder/Encoder`) but then the compiler messages would not point developers to what they're missing in their implementation.

tests/datastore/modifycommontests.nim
benbierens: OK, I understand.
I don't really like it, because this behavior isn't obvious or what a reader would expect from the names. But... I haven't come up with anything better or a different solution. If this is just caused by the TieredDatastore, and we are not using the TieredDatastore anywhere (we aren't), maybe it would be cleaner to drop it entirely?

benbierens (approved):
I approve but there are still useful comments by others that I think should be handled.

tests/datastore/modifycommontests.nim
tbekas: I kind of agree that maybe it would be better to drop the `TieredDatastore` entirely since it's not used anywhere. However such change is not part of the scope for this PR.

tbekas:
I've addressed all the comments above. If any of the reviewers has anything to add please reach out to me.

PR #63, opened 2024-05-14
Adds LevelDB Datastore

benbierens:
Using nim-leveldb static-linked library.

### DIFF
--- a/datastore/leveldb/leveldbds.nim
+++ b/datastore/leveldb/leveldbds.nim
@@ -1,0 +1,178 @@
+import std/options
+import std/tables
+import std/os
+import std/strformat
+
+import pkg/leveldb
+import pkg/chronicles
+import pkg/chronos
+import pkg/questionable
+import pkg/questionable/results
+from pkg/stew/results as stewResults import isErr
+import pkg/upraises
+
+import ../datastore
+import ../datastore/defaultimpl
+
+push: {.upraises: [].}
+
+logScope:
+  topics = "LevelDB"
+
+type
+  LevelDbDatastore* = ref object of Datastore
+    db: LevelDb
+    locks: TableRef[Key, AsyncLock]
+
+func toByteSeq(str: string): seq[byte] {.inline.} =
+  @(str.toOpenArrayByte(0, str.high))
+
+func toString(bytes: openArray[byte]): string {.inline.} =
+  let length = bytes.len
+  if length > 0:
+    result = newString(length)
+    copyMem(result.cstring, bytes[0].unsafeAddr, length)
+
+method has*(self: LevelDbDatastore, key: Key): Future[?!bool] {.async, locks: "unknown".} =
+  try:
+    let str = self.db.get($key)
+    return success(str.isSome)
+  except LevelDbException as e:
+    return failure("LevelDbDatastore.has exception: " & e.msg)
+
+method delete*(self: LevelDbDatastore, key: Key): Future[?!void] {.async, locks: "unknown".} =
+  try:
+    self.db.delete($key, sync = true)
+    return success()
+  except LevelDbException as e:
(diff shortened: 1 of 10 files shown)


datastore/leveldb/leveldbds.nim
```diff
@@ -0,0 +1,157 @@
...
+    locks: TableRef[Key, AsyncLock]
+
+func toByteSeq(str: string): seq[byte] {.inline.} =
+  @(str.toOpenArrayByte(0, str.high))
+
+func toString(bytes: openArray[byte]): string {.inline.} =
+  let length = bytes.len
+  if length > 0:
+    result = newString(length)
+    copyMem(result.cstring, bytes[0].unsafeAddr, length)
```
tbekas: In `stew/byteutils` there are already funcions for this:

```
let 
  bytes = "foo".toBytes()
  str = string.fromBytes(bytes) 
```

datastore/leveldb/leveldbds.nim
```diff
@@ -0,0 +1,157 @@
...
+
+method has*(self: LevelDbDatastore, key: Key): Future[?!bool] {.async, locks: "unknown".} =
```
tbekas: I think that you can remove the `locks` pragma, since the implementation below is known and you know that there are no locks used.

tests/datastore/leveldb/testleveldbds.nim
```diff
@@ -0,0 +1,52 @@
...
+
+  setupAll:
+    createdir(tempDir)
+
+  teardownAll:
+    (await ds.close()).tryGet()
+    removeDir(tempDir)
+
+  basicStoreTests(ds, key, bytes, otherBytes)
+  modifyTests(ds, key)
```
tbekas, line 36: There's a change that just been merged the `TypedDatastore` implementation, which also has it's set of abstract tests. Therefore you can add them as well.

```nim
typedDsTests(ds, key)
```

and in the query section
```nim
typedDsQueryTests(ds)
```

tbekas (approved):
Some minor comments, overall looks good 👍

datastore/leveldb/leveldbds.nim
```diff
@@ -0,0 +1,149 @@
...
+import std/tables
+import std/os
+import std/strformat
+
+import pkg/leveldbstatic
+import pkg/chronos
+import pkg/questionable
+import pkg/questionable/results
+import pkg/stew/byteutils
+from pkg/stew/results as stewResults import isErr
```
dryajov, line 11: Don't use `stew/result` use the standalone results package.

datastore/leveldb/leveldbds.nim
```diff
@@ -0,0 +1,149 @@
...
+  try:
+    let str = string.fromBytes(data)
+    self.db.put($key, str)
+    return success()
+  except LevelDbException as e:
+    return failure("LevelDbDatastore.put exception: " & $e.msg)
+
+method put*(self: LevelDbDatastore, batch: seq[BatchEntry]): Future[?!void] {.async.} =
+  for entry in batch:
+    if err =? (await self.put(entry.key, entry.data)).errorOption:
```
dryajov, line 64: This should be using the batching operations from leveldb.

datastore/leveldb/leveldbds.nim
benbierens: Both fsds and sqliteds also use this same stew package. Do you want those changed as well?

datastore/leveldb/leveldbds.nim
benbierens: Good catch. This was left-over from my initial hacky attempt to plug leveldb into the datastore interface.
Your comments are implemented here: https://github.com/codex-storage/nim-datastore/pull/65/files

PR #67, opened 2024-05-28
Tests for concurrent queries. Finishes disposed iterators.

benbierens:
During debugging, I wrote these tests to check that the datastores can handle concurrent queries.
This didn't help. It didn't reveal any problems. But I figure we might as well keep the tests.

### DIFF
--- a/datastore/fsds.nim
+++ b/datastore/fsds.nim
@@ -190,6 +190,9 @@ method query*(
     iter = QueryIter.new()
 
   proc next(): Future[?!QueryResponse] {.async.} =
+    if iter.finished:
+      return success (Key.none, EmptyBytes)
+
     let
       path = walker()
 
--- a/datastore/fsds.nim
+++ b/datastore/fsds.nim
@@ -216,6 +216,10 @@ method query*(
     return success (key.some, data)
 
   iter.next = next
+  iter.dispose = proc(): Future[?!void] {.async.} =
+    iter.finished = true
+    return success()
+  
   return success iter
 
 method modifyGet*(
--- a/datastore/leveldb/leveldbds.nim
+++ b/datastore/leveldb/leveldbds.nim
@@ -116,6 +116,7 @@ method query*(
 
   proc dispose(): Future[?!void] {.async.} =
     dbIter.dispose()
+    iter.finished = true
     return success()
 
   iter.next = next
--- a/datastore/leveldb/leveldbds.nim
+++ b/datastore/leveldb/leveldbds.nim
@@ -100,7 +100,7 @@ method query*(
 
   proc next(): Future[?!QueryResponse] {.async.} =
     if iter.finished:
-      return failure(newException(QueryEndedError, "Calling next on a finished query!"))
+      return success (Key.none, EmptyBytes)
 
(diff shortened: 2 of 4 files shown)


datastore/fsds.nim
```diff
@@ -216,6 +216,10 @@ method query*(
     return success (key.some, data)
 
   iter.next = next
+  iter.dispose = proc(): Future[?!void] {.async.} =
```
tbekas, line 222: I think that disposing an inter should follow iter being finished, not the other way around.

tbekas (commented):
Nice test 👍

datastore/fsds.nim
benbierens: I'm sorry, I'm not sure what you mean. :o

tbekas: So setting `finished` to true, should be an internal to iterator event that's signaling that all items were yielded (line 197).

With this change, flag `finished` would be set to true whenever `dispose()`  would be called (even if all items were not yielded). So as a client of this API you can finish iteration prematurely and call `dispose()`, but that's not the same as yielding all elements.

gmega: :red_circle: Well the client won't even finish iteration because `next` will still happily return the next element as `next` doesn't check the `finished` flag to decide what to do. So if you're relying on it returning `success (Key.none, ...)` to know when to stop, this will not have the intended effect.

I suppose that to get the right semantics (assuming this is the right one) you need to check the state of the finished flag in addition to the state of the walker iterator (e.g `if finished(walker) or iter.finished`).

gmega (changes requested):
The tests are fine, but the semantics for disposal is a bit weird for fsds, wanna make sure that's what's intended.

datastore/fsds.nim
benbierens: I think I understand the desired behavior and I'm gonne write a test for it.

benbierens: Update: Finished iters will always return None when asked for next. There's a test to ensure this, and ensure that disposed iters become finished.

dryajov: Not sure why we're (incorrectly) changing the semantics of the iterators now? 

as @tbekas is saying:

> So setting finished to true, should be an internal to iterator event that's signaling that all items were yielded (line 197).

gmega: I think the goal is to allow dispose to be called before the iterator returns all elements. I suppose this is useful because it's not always that you will want to consume the whole set of results from a query, and in that case you'll want to free up the resources (dispose) anyway. Is that the use case @benbierens?

gmega: Yeah OK the concern I had is fixed by this.

gmega (approved):
OK the concerns I had have been addressed so approving it.

datastore/fsds.nim
benbierens: Yes, we can't assume the iterator will always run through all elements. I've added a test to show that you can properly clean up an iterator early by calling dispose on it. Any nexts would then yield nothing and the underlying resources can be closed.

PR #68, opened 2024-06-04
Rename err to error to avoid naming conflict

tbekas:
Label `err` in generic procs (which are basically templates) often conflicts with this definition from `nim-results`: `template err*(): auto = err(typeof(result))`. Which causes often unexpected compilation errors.


benbierens (approved):
what could possibly go wrong!?

gmega (approved):
LGTM

gmega:
It's complaining that you haven't signed your commits, @tbekas (or is it really you??)

PR #70, opened 2024-08-14
Skip unnecessary writes in default impl for modifyGet

tbekas:
This is an optimization only. No functionality is changed, therefore no need to update any tests.

### DIFF
--- a/datastore/defaultimpl.nim
+++ b/datastore/defaultimpl.nim
@@ -37,6 +37,10 @@ proc defaultModifyGetImpl*(
     except CatchableError as err:
       return failure(err)
 
+    if maybeCurrentData == maybeNewData:
+      # no need to change currently stored value (if any)
+      return aux.success
+
     if newData =? maybeNewData:
       if err =? (await self.put(key, newData)).errorOption:
         return failure(err)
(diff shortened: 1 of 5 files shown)


benbierens (dismissed):
👍

datastore/defaultimpl.nim
```diff
@@ -37,6 +37,10 @@ proc defaultModifyGetImpl*(
     except CatchableError as err:
       return failure(err)
 
+    if maybeCurrentData == maybeNewData:
```
gmega, line 40: :red_circle: Well technically you should be able to test for this (the absence of a write), and I'd argue you should test for it as otherwise you can't know if your optimization is actually working (which is something we care about, as otherwise there'd be no point in putting it there), no matter how convincing the code looks. :slightly_smiling_face:

I'm wiling to let this pass - against my own best judgement - if you tell me that writing such a test is complicated.

gmega (changes requested):

datastore/defaultimpl.nim
tbekas: I think that skipping writes is not such a crucial feature that would justify having a test that verify if something was called or not called (one of the worst kinds of tests imo), but I've added such test anyway.

benbierens (approved):
I like it

PR #71, opened 2024-08-19
Simplifies tests.yml

benbierens:
This is done in response to choosenim's download URL breaking and causing the CI to fail.
Mark suggested replacing choosenim with a iffy nim-specific github action. This is already being used in several other repositories. So this PR rolls out the same solution.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,19 +1,9 @@
 name: Tests
-
-on:
-  push:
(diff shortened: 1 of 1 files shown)


veaceslavdoina (dismissed):

gmega (approved):
Oh I like this! :grin:

veaceslavdoina (approved):

PR #73, opened 2024-08-26
Sets proper version for chronos

benbierens:
The hash-reference to chronos is breaking dependency resolution in nim-codex-dht.
This should fix that.

### DIFF
--- a/datastore.nimble
+++ b/datastore.nimble
@@ -1,14 +1,14 @@
 mode = ScriptMode.Verbose
 
 packageName   = "datastore"
-version       = "0.1.0"
+version       = "0.1.1"
 author        = "Status Research & Development GmbH"
 description   = "Simple, unified API for multiple data stores"
 license       = "Apache License 2.0 or MIT"
 
 requires "nim >= 1.2.0",
          "asynctest >= 0.5.2 & < 0.6.0",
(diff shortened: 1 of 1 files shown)


gmega (approved):
That's super weird cause [the nim-codex-dht hash is actually the same](https://github.com/codex-storage/nim-codex-dht/blob/5f38fd95703feb36f3cae76c037f6e96192a07a7/codexdht.nimble#L15) so it should be able to resolve it. But OK pinning hashes in the nimble file is bad anyways so to me this looks better.

veaceslavdoina (approved):

dryajov:
Why don't we have a lock file? I though we added this at some point? Without a lock file, you can't guarantee that you are on the correct version as nimble will try to use a (broken) semver heuristic to pull in random stuff 🤦

dryajov (approved):
But we should really add a lock file

gmega:
Lock files won't save you from broken semver because whenever you need to update a dep that's what it'll go for. Instead you'll end up having to edit them manually, which involves updating fields which were simply not meant to be updated that way. It's easier to use submodules (e.g. nimbus-build-system) at that point.

benbierens:
It seems to be true that if we consistently apply these version-range identifications in the nimble files across repositories, it works without any trouble. This is what Mark's been recommending me, and I'm trying to get everything consistent in our repos.

PR #77, opened 2025-02-25
chore: define raises for async pragma

2-towns:
This PR adds `raises` to the async pragma.

### DIFF
--- a/datastore/tieredds.nim
+++ b/datastore/tieredds.nim
@@ -47,8 +47,11 @@ method delete*(
     pending = await allFinished(self.stores.mapIt(it.delete(key)))
 
   for fut in pending:
-      without res =? fut.read().catch, error:
-        return failure error
+    try:
+      if fut.read().isErr:
+        return fut.read()
+    except FuturePendingError as err:
+      return failure err
 
   return success()
 
@@ -61,8 +64,11 @@ method delete*(
       pending = await allFinished(self.stores.mapIt(it.delete(key)))
 
     for fut in pending:
-      without res =? fut.read().catch, error:
-        return failure error
+      try:
+        if fut.read().isErr:
+          return fut.read()
+      except FuturePendingError as err:
+        return failure err
 
   return success()
 
@@ -99,8 +105,11 @@ method put*(
     pending = await allFinished(self.stores.mapIt(it.put(key, data)))
 
   for fut in pending:
-    without res =? fut.read().catch, error:
-      return failure error
+    try:
+      if fut.read().isErr:
+        return fut.read()
+    except FuturePendingError as err:
+      return failure err
 
   return success()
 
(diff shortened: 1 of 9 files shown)


datastore/tieredds.nim
```diff
@@ -41,32 +41,40 @@ method has*(
...
   let
     pending = await allFinished(self.stores.mapIt(it.delete(key)))
 
   for fut in pending:
-    if fut.read().isErr: return fut.read()
+    try:
+      if fut.read().isErr:
+        return fut.read()
+    except FuturePendingError as err:
+      return failure err
```
emizzle: For all of these blocks below, I don't think we really need to read the future, we can simply await it, and since they don't raise exceptions besides `CancelledError`, we don't need a try/except, eg:
```nim
for fut in pending:
  let res = await fut
  if res.isErr: return res
```

emizzle (dismissed):
Nice job. My suggestion is definitely not blocking, so will approve now.

datastore/tieredds.nim
```diff
@@ -47,11 +47,8 @@ method delete*(
     pending = await allFinished(self.stores.mapIt(it.delete(key)))
 
   for fut in pending:
-    try:
-      if fut.read().isErr:
-        return fut.read()
-    except FuturePendingError as err:
-      return failure err
+    let res = await fut
+    if res.isErr: return res
```
emizzle: Sorry, I thought about this super late last night. We can probably simplify this even further with:
```suggestion
    ? await fut
```

The `?` is a nim-results operator that will return the value (if there is one) or return the Result error otherwise. My  apologies for not thinking of this yesterday!

emizzle:
```nim
import pkg/results
import pkg/chronos

proc underlyingDelete(isError = false): Future[Result[void, int]] {.async.} =
  if isError:
    return err 1
  else:
    return ok()

proc delete(isError: bool): Future[Result[void, int]] {.async.} =
  ?await underlyingDelete(isError)
  return ok()

let e = waitFor delete(true)
let k = waitFor delete(false)
assert e.isErr
assert e.error == 1
assert k.isOk
```

emizzle (dismissed):
Sorry, I only thought of a better way to do what I had suggested late last night. Definitely not a blocker, so will approve now.

datastore/tieredds.nim
2-towns: No problem.

When there is a value, it is not really `returning` but more like `assigning` the value if a variable is declared like the following line, right ?

```nim
let res = ?await underlyingDelete(isError)
```

So, what I understand is that when there is a value, the function will proceed, but if there is an error, the function will return that error. Am I right?

I tried to implement this in my last commit; please double-check those two lines, particularly [here](https://github.com/codex-storage/nim-datastore/blob/chore/define-errors-raised-in-async-pragam/datastore/tieredds.nim#L128-L129).

emizzle: Yes, that's correct. If there's a value, it will be pulled out of the Result (in the case of `Result[void, xx]` there is no value, just `void`). If there's an error, the current scope will return with an errored Result returned from the function call.

emizzle: Those two lines look good, well done 👍

AuHau (approved):
LGTM

datastore/tieredds.nim
2-towns: Ok thanks !

arnetheduck (approved):

PR #80, opened 2025-11-06
chore: update stew upper bound version

cnanakos:
Bump package version to 0.2.1.

### DIFF
--- a/datastore.nimble
+++ b/datastore.nimble
@@ -1,7 +1,7 @@
 mode = ScriptMode.Verbose
 
 packageName   = "datastore"
-version       = "0.2.0"
+version       = "0.2.1"
 author        = "Status Research & Development GmbH"
 description   = "Simple, unified API for multiple data stores"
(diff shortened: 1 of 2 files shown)


datastore.nimble
```diff
@@ -12,7 +12,7 @@ requires "nim >= 2.0.14",
          "questionable >= 0.10.15 & < 0.11.0",
          "sqlite3_abi >= 3.47.0.0 & < 4.0.0.0",
          "leveldbstatic >= 0.2.0 & < 0.3.0",
-         "stew >= 0.2.0 & < 0.3.0",
+         "stew >= 0.2.0 & < 0.5.0",
```
2-towns: I would suggest to increase the lower bound version as well to make sure that you use the last version of stew.

cnanakos: Updated to 0.4.0 as suggested.

2-towns (approved):

arnetheduck (approved):

PR #83, opened 2026-02-05
fix: close pending iterators before closing LevelDB store

### LINKED ISSUE #82: LevelDB datastore crashes if one attempts to close it while there are still iterators which haven't been disposed of
The description says it. If you try to do that, the program will crash with:

```text
leveldb::VersionSet::~VersionSet(): Assertion `dummy_versions_.next_ == &dummy_versions_' failed.
```

This is particularly common when shutting down the node.

gmega:
This PR adds tracking of open iterators to leveldbds so that when one attempts to close it, iterators are disposed of first. It also adds automatic disposal if iterators are completely consumed.

fixes #82

### DIFF
--- a/datastore/leveldb/leveldbds.nim
+++ b/datastore/leveldb/leveldbds.nim
@@ -98,6 +98,11 @@ method query*(
       limit = query.limit
     )
 
+  proc dispose(): Future[?!void] {.async: (raises: [CancelledError]).} =
+    dbIter.dispose()
+    iter.disposed = true
+    return success()
+
   proc next(): Future[?!QueryResponse] {.async: (raises: [CancelledError]).} =
     if iter.finished:
       return failure(newException(QueryEndedError, "Calling next on a finished query!"))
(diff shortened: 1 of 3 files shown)


datastore/leveldb/leveldbds.nim
```diff
@@ -98,6 +107,13 @@ method query*(
       limit = query.limit
     )
 
+  proc dispose(): Future[?!void] {.async: (raises: [CancelledError]).} =
+    dbIter.dispose()
+    iter.disposed = true
+    self.openIterators.excl(iter)
```
emizzle, line 113: Possible optimisation: check `iter.disposed` first?

emizzle (approved):

datastore/leveldb/leveldbds.nim
gmega: Those are so fast and the operation so infrequent that's probably not even worth the trouble... (plus dbIter.dispose is idemptotent)

